Issue #11, opened 2024-07-08
Clean up issues in `astro check`


fengelniederhammer:
`npm run check-astro` shows a lot of errors. Clean them up so that we can integrate the command in the CI.

PR #142, opened 2024-09-19
refactor: unify organism names and refactor their config

### LINKED ISSUE #109: Notifications: Revisit naming of instance names
The currently occur in `Subscriptions.kt` and `SubscriptionTable.tsx`. They have different casing. Unify them

fengelniederhammer:
resolves #109

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/subscriptions/Subscription.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/subscriptions/Subscription.kt
@@ -14,16 +14,20 @@ enum class DateWindow {
 enum class Organism {
     @JsonProperty("covid")
     COVID,
+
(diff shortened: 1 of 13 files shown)


JonasKellerer (approved):
LGTM.
I am interessed where this will go: if we need to add more to the config in the future, or how/if we will set the specific values of colors for the different organisms.

JonasKellerer (changes requested):
I think you missed in `src/pages/subscriptions/index.astro` the dummy values on the website. For me `http://localhost:4321/subscriptions` shows an error.

fengelniederhammer:
> I think you missed in `src/pages/subscriptions/index.astro`

Thanks for catching this. I just checked: `npm run check-astro` would have caught it, but we don't run it in CI, since it throws too many errors. We should do it now... #11

JonasKellerer (approved):
LGTM

PR #143, opened 2024-09-20
80 notification backend add count notification computation logic

### LINKED ISSUE #80: Notification backend: add count notification computation logic
Read the dashboards_config.json. It contains the information which LAPIS instance to call for which organism.

Implement the trigger logic for the "count" trigger:
The trigger contains a window to look at (e.g. last two months)
Compute the dateTo and dateFrom depending on today's date
Call the respective LAPIS instance with the trigger filter and the computed dateTo and dateFrom.

The subscription should contain the information which date column to look at.

fengelniederhammer:
resolves #80

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/TriggerEvaluation.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/TriggerEvaluation.kt
@@ -1,0 +1,33 @@
+package org.genspectrum.dashboardsbackend.api
+
+data class TriggerEvaluationResponse(
+    val result: TriggerEvaluationResult,
+)
+
+sealed interface TriggerEvaluationResult {
+    data class EvaluationError(
+        val message: String,
+    ) : TriggerEvaluationResult {
+        val type = EvaluationErrorType.EvaluationError
+
+        enum class EvaluationErrorType {
+            EvaluationError,
+        }
+    }
+
+    data object ConditionMet : TriggerEvaluationResult {
+        val type = ConditionMetType.ConditionMet
+
+        enum class ConditionMetType {
+            ConditionMet,
+        }
+    }
+
+    data object ConditionNotMet : TriggerEvaluationResult {
+        val type = ConditionNotMetType.ConditionNotMet
+
+        enum class ConditionNotMetType {
+            ConditionNotMet,
+        }
+    }
+}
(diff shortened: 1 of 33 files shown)


backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/TriggerEvaluation.kt
```diff
@@ -16,15 +16,23 @@ sealed interface TriggerEvaluationResult {
         }
     }
 
-    data object ConditionMet : TriggerEvaluationResult {
+    data class ConditionMet(
+        val actualResult: Int,
```
JonasKellerer, line 20: I'm not happy with the naming. How about `evaluatedValue`?

backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/SubscriptionsControllerTriggerEvaluationTest.kt
```diff
@@ -0,0 +1,107 @@
...
+    @Autowired private val subscriptionsClient: SubscriptionsClient,
+) {
+    @MockkBean
+    private lateinit var lapisClientProviderMock: LapisClientProvider
+
+    @Test
+    fun `GIVEN lapis returns count greater than threshold WHEN evaluating trigger THEN returns condition met`() {
+        val userId = getNewUserId()
+
+        val createdSubscription = subscriptionsClient.postSubscription(dummySubscriptionRequest, userId)
```
JonasKellerer, line 39: No serious complaint, but can we somehow overwrite the value of the trigger thershold here from the dummySubscriptionRequest through a copy, or something similar? Only from reading the test it is not clear, why this test should pass and the dummyRequest lies in another file.

backend/src/main/resources/db/migration/V1__init.sql
```diff
@@ -3,10 +3,8 @@ create table subscriptions_table (
     name text not null,
     interval VARCHAR(255) not null,
     active boolean not null,
-    conditions_met boolean not null,
```
JonasKellerer, line 6: Just a comment: When we do it this, we need to call in the frontend the `/evaluateSubscription` endpoint for each subscription the user has on the `/subscriptions` frontend page. Is this what we want to do? Should the user be shown the current values, or the stored values of the trigger with an additional button to evaluate the triggers?

JonasKellerer (approved):
Nice work.
I think we need to discuss how the subscriptions should be evaluated in the frontend. This way one needs to call for each subscription on the subscription page the endpoint and get the result.

backend/src/main/resources/db/migration/V1__init.sql
fengelniederhammer: As discussed: We can still add it again later. Latest when we implement notifications via some channel, we probably need to store the last result anyway.

PR #195, opened 2024-10-08
Make sure the plots only use latest versions from lapis

### LINKED ISSUE #141: Make sure the plots only use "latest versions" from LAPIS
We have some LAPIS instances that are backed by a Loculus instance. The ingress potentially generates new versions (revisions) of some accessions. We need to make sure that we filter out non-latest versions properly:
* Either by setting LAPIS filters accordingly
  * Advantage: Easy to implement
  * Disadvantage: All users of the LAPIS instances need to be aware that this is necessary.
* Or by filtering out non-latest versions before they get into SILO
  * Advantage: client-friendly - no user needs to think about Loculus interna (which we actually want to hide)
  * Disadvantage: maybe a bit tedious in implementation, as we need to modify the link between Loculus backend and SILO preprocessing.

JonasKellerer:
resolves #141

### Summary

Each time LAPIS is called for WestNile an additional filter for the latest version is applied. This is necessary, because Loculus provides the whole data, including revisions, to LAPIS.

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/triggerevaluation/TriggerEvaluator.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/triggerevaluation/TriggerEvaluator.kt
@@ -3,6 +3,7 @@ package org.genspectrum.dashboardsbackend.model.triggerevaluation
 import kotlinx.datetime.DateTimeUnit
 import kotlinx.datetime.minus
 import org.genspectrum.dashboardsbackend.api.DateWindow
+import org.genspectrum.dashboardsbackend.api.Organism
 import org.genspectrum.dashboardsbackend.api.Subscription
 import org.genspectrum.dashboardsbackend.api.Trigger
 import org.genspectrum.dashboardsbackend.api.TriggerEvaluationResult
@@ -22,20 +23,21 @@ class TriggerEvaluator(
         val lapisClient = lapisClientProvider.provide(subscription.organism)
 
         val dateFilter = computeDateFilter(subscription)
+        val additionalFilters = additionalFilters(subscription.organism)
 
         val computation = when (subscription.trigger) {
             is Trigger.CountTrigger -> CountComputation(
                 subscription = subscription,
                 lapisClient = lapisClient,
-                lapisFilter = subscription.trigger.filter + dateFilter,
+                lapisFilter = subscription.trigger.filter + dateFilter + additionalFilters,
                 threshold = subscription.trigger.count,
             )
 
             is Trigger.ProportionTrigger -> ProportionComputation(
                 subscription = subscription,
                 lapisClient = lapisClient,
-                numeratorFilter = subscription.trigger.numeratorFilter + dateFilter,
-                denominatorFilter = subscription.trigger.denominatorFilter + dateFilter,
+                numeratorFilter = subscription.trigger.numeratorFilter + dateFilter + additionalFilters,
+                denominatorFilter = subscription.trigger.denominatorFilter + dateFilter + additionalFilters,
                 threshold = subscription.trigger.proportion,
             )
         }
@@ -56,6 +58,10 @@ class TriggerEvaluator(
             "${dateField}From" to "$lowerBoundDate",
             "${dateField}To" to "$currentDate",
         )
+    }
+
+    private fun additionalFilters(organism: Organism): Map<String, String> {
+        return dashboardsConfig.getOrganismConfig(organism).lapis.additionalFilters ?: emptyMap()
     }
 }
 
--- a/website/src/components/rsv/RsvSequencingEffortsPage.astro
+++ b/website/src/components/rsv/RsvSequencingEffortsPage.astro
@@ -37,7 +37,7 @@ const { label: subdivisionLabel, field: subdivisionField } = getLocationSubdivis
             <div class='max-w-[1000px]'>
                 <LocationTimeFilter
                     fields={view.locationFields}
-                    initialLocation={baselineFilter}
+                    initialLocation={routeData.baselineFilter.location}
                     initialDateRange={routeData.baselineFilter.dateRange}
                     earliestDate={view.earliestDate}
                     customDateRangeOptions={view.customDateRangeOptions}
--- a/website/src/routes/covid.ts
+++ b/website/src/routes/covid.ts
@@ -20,14 +20,13 @@ import { organismConfig, Organisms } from '../types/Organism.ts';
 const pathFragment = organismConfig[Organisms.covid].pathFragment;
 
 const earliestDate = '2020-01-06';
-const today = getTodayString();
 
 class CovidConstants {
     public readonly organism = Organisms.covid;
     public readonly defaultDateRange: DateRange = 'last6Months';
     public readonly earliestDate = '2020-01-06';
     public readonly customDateRangeOptions: DateRangeOption[] = [
-        { label: '2024', dateFrom: '2024-01-01', dateTo: today },
+        { label: '2024', dateFrom: '2024-01-01', dateTo: getTodayString() },
         { label: '2023', dateFrom: '2023-01-02', dateTo: '2023-12-31' },
         { label: '2022', dateFrom: '2022-01-03', dateTo: '2023-01-01' },
         { label: '2021', dateFrom: '2024-01-04', dateTo: '2022-01-02' },
@@ -39,6 +38,7 @@ class CovidConstants {
     public readonly hostField: string;
     public readonly originatingLabField: string | undefined;
     public readonly submittingLabField: string | undefined;
+    public readonly additionalFilters: Record<string, string> | undefined;
 
     constructor(organismsConfig: OrganismsConfig) {
         this.mainDateField = organismsConfig.covid.lapis.mainDateField;
@@ -47,6 +47,7 @@ class CovidConstants {
         this.hostField = organismsConfig.covid.lapis.hostField;
         this.originatingLabField = organismsConfig.covid.lapis.originatingLabField;
         this.submittingLabField = organismsConfig.covid.lapis.submittingLabField;
+        this.additionalFilters = organismsConfig.covid.lapis.additionalFilters;
     }
 
     public variantFilterToLapisFilter(filter: LapisCovidVariantQuery): LapisFilter {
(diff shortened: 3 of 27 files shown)


backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/triggerevaluation/TriggerEvaluator.kt
```diff
@@ -57,6 +59,13 @@ class TriggerEvaluator(
             "${dateField}To" to "$currentDate",
         )
     }
+
+    private fun getLatestVersionFilter(organism: Organism): Map<String, String> {
+        if (organism != Organism.WestNile) {
+            return emptyMap()
+        }
+        return mapOf("versionStatus" to "LATEST_VERSION")
```
chaoran-chen, line 67: It should also filter for `isRevocation=false`

JonasKellerer: Thanks for catching that. I had a small voice in my head saying that I am missing something, but I could not find out what.

JonasKellerer: I fixed it.

fengelniederhammer (changes requested):
I'd prefer to have the filters defined in the config. I imagine something like
```
    h5n1:
      lapis:
        url: "https://lapis.genspectrum.org/h5n1"
        mainDateField: "sample_collection_date"
        locationFields:
          - "geo_loc_country"
          - "geo_loc_admin_1"
        lineageField: "lineage"
        hostField: "host_name_scientific"
        authorAffiliationsField: "author_affiliations"
        authorsField: "authors"
        additionalFilters:
            versionStatus: "LATEST_VERSION"
            isRevocation: false
```

website/src/components/covidCompareVariantsView/JointFilter.astro
```diff
@@ -90,6 +91,7 @@ const initialDateRangeTo = isCustomDateRange(initialDateRange) ? initialDateRang
             super();
             let { filterId, location, dateRange, lineage, mutations } = JSON.parse(this.dataset.message!);
             const config = JSON.parse(this.dataset.organismsConfig!);
+            const additionalFilters = JSON.parse(this.dataset.additionalFilters!);
```
fengelniederhammer, line 94: It already has the whole config anyway, we can simply use it:
```suggestion
            const additionalFilters = config.covid.lapis.additionalFilters
```

website/src/components/subscriptions/create/TriggerInput.tsx
```diff
@@ -7,6 +7,7 @@ import { BaselineInput } from './BaselineInput.tsx';
 import { VariantInput } from './VariantInput.tsx';
 import type { DateWindow } from '../../../types/DateWindow.ts';
 
+// TODO: Add latest version filter for WestNile #194
```
fengelniederhammer, line 10: Do we really need to do something here? I would have expected that we only need to change something in `FilterDisplay`.

website/src/components/covidCompareVariantsView/JointFilter.astro
```diff
@@ -125,8 +127,13 @@ const initialDateRangeTo = isCustomDateRange(initialDateRange) ? initialDateRang
                             baselineFilter: {
                                 location,
                                 dateRange,
+                                ...additionalFilters,
+                            },
+                            variantFilter: {
+                                lineage,
+                                ...mutations,
+                                ...additionalFilters,
```
fengelniederhammer, line 135: Actually I think this should not need the additional filters at all...

fengelniederhammer (changes requested):
Something is weird about the URLs:

* go to `/west-nile/sequencing-efforts`
* Select country Albania and submit filter
* the URL will be `/west-nile/sequencing-efforts?geoLocCountry=Albania`
* select "last 2 weeks" in date filter and submit
* the URL will be `/west-nile/sequencing-efforts?geoLocCountry=Albania&sampleCollectionDateFrom=1930-01-01&sampleCollectionDateTo=2024-10-14&versionStatus=LATEST_VERSION&isRevocation=false&sampleCollectionDate=2024-09-30--2024-10-14`
* change country to Australia and submit
* the URL will be `/west-nile/sequencing-efforts?geoLocCountry=Australia&sampleCollectionDate=2024-09-30--2024-10-14`

I think the additional filters should not be part of the URL in general. A user should not need to see them or be able to change them.

website/src/components/subscriptions/create/TriggerInput.tsx
JonasKellerer: True. Initially, I thought to change the filters here to also store them in the backend. Now we don't store them. I moved this comment to `FilterDisplay`

website/src/components/rsv/RsvSequencingEffortsPage.astro
```diff
@@ -37,7 +37,7 @@ const { label: subdivisionLabel, field: subdivisionField } = getLocationSubdivis
             <div class='max-w-[1000px]'>
                 <LocationTimeFilter
                     fields={view.locationFields}
-                    initialLocation={baselineFilter}
+                    initialLocation={routeData.baselineFilter.location}
```
fengelniederhammer, line 40: This is confusing:

I was wondering why we don't simply use `baselineFilter.location` instead of `routeData.baselineFilter.location`, but those are completely different types and different concepts.
* The `baselineFilter` is the object that is prepared to be sent as a LAPIS filter
* `routeData.baselineFilter` is the internal state of the website that can be transformed into the `baselineFilter`.

I think we should separate those in terminology to make the distinction clearer between "LAPIS filter" and "dashboard state".

Definitely not something to do in this PR, but to keep in mind.

website/src/routes/covid.ts
```diff
@@ -298,22 +317,25 @@ export class CovidCompareVariantsView
             ...filter.location,
             [`${this.mainDateField}From`]: dateRange.from,
             [`${this.mainDateField}To`]: dateRange.to,
+            ...this.additionalFilters,
         };
     }
 
-    private getFilter(filterParams: Map<string, string>) {
+    private readonly getFilter = (filterParams: Map<string, string>) => {
```
fengelniederhammer, line 324: Let's leave this a proper function instead of an arrow function?

fengelniederhammer (changes requested):

website/src/components/rsv/RsvSequencingEffortsPage.astro
fengelniederhammer: But there is also an issue (from the same confusion) with this PR:

I think we only need the `additionalFilters` in the `View.toLapisFilter` methods, as those represent the transition from "dashboard state" to "LAPIS filter". We don't need the `additionalFilters` in the "dashboard state", since it's already in the config, i.e. we need it neither in `View.parseUrl` functions nor in the `View.defaultRoute`s.

JonasKellerer: True. I changed it, so the additionalFilters are only applied during conversion to a lapisFilter. 
I also agree, that the naming is confusing. I added a ticket for renaming #222

website/src/routes/covid.ts
```diff
@@ -92,7 +95,9 @@ export class CovidAnalyzeSingleVariantView extends CovidConstants implements Vie
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisCovidVariantQuery(search, this.lineageField),
+            variantFilter: {
+                ...getLapisCovidVariantQuery(search, this.lineageField),
+            },
```
fengelniederhammer, line 100: Let's revert this?

website/src/routes/h5n1.ts
```diff
@@ -79,33 +82,35 @@ export class H5n1AnalyzeSingleVariantView extends H5n1Constants implements View<
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisVariantQuery(search, this.lineageField),
+            variantFilter: {
+                ...getLapisVariantQuery(search, this.lineageField),
+            },
```
fengelniederhammer, line 87: Let's revert this?

website/src/routes/mpox.ts
```diff
@@ -92,33 +95,35 @@ export class MpoxAnalyzeSingleVariantView extends MpoxConstants implements View<
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisVariantQuery(search, this.lineageField, this.cladeField),
+            variantFilter: {
+                ...getLapisVariantQuery(search, this.lineageField, this.cladeField),
+            },
```
fengelniederhammer, line 100: Let's revert this?

website/src/routes/rsvA.ts
```diff
@@ -79,33 +82,35 @@ export class RsvAAnalyzeSingleVariantView extends RsvAConstants implements View<
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisVariantQuery(search, this.lineageField),
+            variantFilter: {
+                ...getLapisVariantQuery(search, this.lineageField),
+            },
```
fengelniederhammer, line 87: Let's revert this?

website/src/routes/rsvB.ts
```diff
@@ -79,33 +82,35 @@ export class RsvBAnalyzeSingleVariantView extends RsvBConstants implements View<
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisVariantQuery(search, this.lineageField),
+            variantFilter: {
+                ...getLapisVariantQuery(search, this.lineageField),
+            },
```
fengelniederhammer, line 87: Let's revert this?

website/src/routes/westNile.ts
```diff
@@ -80,33 +79,35 @@ export class WestNileAnalyzeSingleVariantView extends WestNileConstants implemen
                 location: getLapisLocationFromSearch(search, this.locationFields),
                 dateRange: getDateRangeFromSearch(search, this.mainDateField) ?? this.defaultDateRange,
             },
-            variantFilter: getLapisVariantQuery(search, this.lineageField),
+            variantFilter: {
+                ...getLapisVariantQuery(search, this.lineageField),
+            },
```
fengelniederhammer, line 84: Let's revert this?

fengelniederhammer (approved):
LGTM :+1:

PR #196, opened 2024-10-08
Add error page

### LINKED ISSUE #135: Add error page
Currently, when something goes wrong, the Astro server will simply return 500 and an empty page (in production mode). Configure an error page, such that users at least see something like "Oops, our fault".

JonasKellerer:
resolves #135

### Summary

Creates both a custom 404 and 500 error page.

### DIFF
--- a/website/src/pages/500.astro
+++ b/website/src/pages/500.astro
@@ -1,0 +1,18 @@
+---
+import BaseLayout from '../layouts/base/BaseLayout.astro';
+import { PageContainer } from '../styles/containers/PageContainer';
+import { PageHeadline } from '../styles/containers/PageHeadline';
+---
+
+<BaseLayout title='GenSpectrum Dashboards' omitHeaderNavigation={true}>
+    <PageContainer>
+        <PageHeadline>Oops! An error occurred...</PageHeadline>
+        <div class='mb-2'>We are sorry for the inconvenience. Please try again.</div>
+        <div>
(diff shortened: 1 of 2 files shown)


website/src/pages/500.astro
fengelniederhammer, line 1: I'm thinking whether we should add some more details to this page. If users are supposed to create an issue, it's valuable for us to know what actually went wrong.

https://docs.astro.build/en/basics/astro-pages/#error
Maybe we can extract something from the error prop?

An alternative would be logging the error prop and showing the user some tracing info that we can use to retrieve the error from logs. That could be a request id or an "error id" (a UUID that we generate here and log it with the error). Anyway, we need
* to tell users: "copy paste *this* to the issue so that we can reproduce the issue"
* a way to retrieve the error information using *this*

It's probably worth it's own issue (along with proper logging in general).

fengelniederhammer (approved):

website/src/pages/500.astro
JonasKellerer: I think it would be good to show more information to the user on this page. For that we need more tooling, namely more diverse error types. Currently, we only have a generic error from the backend.

I would hessitate to provide users with errors from the backend direclty. I think they are of not much use for the users, since they only want to use our page and might be confused by technical details. It would be nice, to show them something they can do to fix that error. However, it seems to be hard to do. 

I like the idea of the UUID message. I created a ticket for that #201.

PR #199, opened 2024-10-08
feat(website): show trigger condition and notification status separately #180

### LINKED ISSUE #180: Separate "Active" and "Condition met" indicator
On the subscription overview page, have a separate indicator if a subscription is inactive, meaning the user does not want to get notified when the condition is met via slack/email/etc, and an indicator that the condition is met.

JonasKellerer:
resolves #180

### Summary
This separates the notification status and trigger condition icons. The notification status is now shown next to the notification channels.

### Screenshot

grafik


website/src/components/subscriptions/overview/SubscriptionEntry.tsx
```diff
@@ -26,8 +22,9 @@ export function SubscriptionEntry({ subscription }: { subscription: Subscription
                             <CardDescription
                                 title={subscription.name}
                                 subtitle={organismConfig[subscription.organism].label}
-                                icon={`size-6 ${getIcon()}`}
+                                icon={`size-8 ${getConditionIcon()}`}
                             />
```
fengelniederhammer, line 26: As discussed - a tooltip would be nice that explains the icon.

fengelniederhammer (approved):
LGTM

As discussed - we can think about the status display icons later and maybe choose other icons if we find more suitable ones.

website/src/components/subscriptions/overview/SubscriptionEntry.tsx
JonasKellerer: grafik

Issue #211, opened 2024-10-14
Round evaluated value


JonasKellerer:
On the subscription overview page, round the evaluated value to a meaningful value. Currently, it looks like this

![Bildschirmfoto vom 2024-10-14 08-58-55](https://github.com/user-attachments/assets/8bfe29fd-f9fc-4d88-bf22-04019ab87376)

A possible solution could be, that it is rounded to one digit more than the last digit of the threshold.

PR #232, opened 2024-10-23
feat(website): round trigger evaluation results (#211)

anna-parker:
resolves https://github.com/GenSpectrum/dashboards/issues/211

### Summary

Round to precision 3 decimal points:
image
 otherwise for evaluated values below threshold: 
image
keep integers as they are:
image

### DIFF
--- a/website/src/components/subscriptions/overview/SubscriptionDisplay.tsx
+++ b/website/src/components/subscriptions/overview/SubscriptionDisplay.tsx
@@ -56,6 +56,26 @@ function DateWindowDisplay({ dateWindow }: { dateWindow: DateWindow }) {
     );
 }
 
+function roundToPrecision({
+    val,
+    precision = 3,
+    lowerBound = 0.001,
+}: {
+    val: number;
+    precision?: number;
+    lowerBound?: number;
+}): string {
+    if (val < lowerBound) {
+        return `<${lowerBound}`;
+    }
+
+    if (Number.isInteger(val)) {
+        return val.toString();
+    }
+
+    return val.toFixed(precision);
+}
+
 function TriggerEvaluationResult({ result }: { result: Subscription['triggerEvaluationResult'] }) {
     if (result.type === 'EvaluationError') {
         return (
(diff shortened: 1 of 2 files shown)


website/src/components/subscriptions/overview/SubscriptionDisplay.tsx
```diff
@@ -73,7 +93,7 @@ function TriggerEvaluationResult({ result }: { result: Subscription['triggerEval
                     {result.type === 'ConditionMet' ? 'met' : 'not met'}
                 </span>
             </div>
-            <div className='text-gray-500'>Evaluated value: {result.evaluatedValue}</div>
+            <div className='text-gray-500'>Evaluated value: {roundToPrecision({ val: result.evaluatedValue })}</div>
```
JonasKellerer, line 96: When I set the thershold to a value below 0.001, then I always get the "<" sign, even when the trigger condition is met.

I got two ideas, but I'm not sure how much effort they are in respect to the benefit:
1) Just show the value, rounded to the same precission as the thershold, when the evaluated value is above the thersohld.
2) Set the lower bound to 0.1% of the thershold, rounded to the next decimal. So a thershold of 0.0003 would have a lower bound of 0.0000001.

@chaoran-chen Does this even make sense to set that low thresholds?

chaoran-chen: Difficult to say for sure but I can't see a good use case right way when a threshold below 0.1% would be useful, so I'm happy with defining this as the lowest allowed threshold.

JonasKellerer: As discussed: We leave it like this. This means, that when the user inserts a very low threshold, then even when the evaluated value is larger than this threshold but still lower than the defined precision it is truncated to "<0.1%".
There are now two new tickets related: #244, #245.

JonasKellerer (approved):
Nice. Thank you. :)

fengelniederhammer:
Should we maybe display the values as percent instead of plain floats? (i.e. `15%` instead of `0.15`)

JonasKellerer:
The component now shows the evaluated value and the threshold with a % sign, when the trigger is set to proportion:
grafik

fengelniederhammer (approved):
It's a bit weird that it now also shows `100.000%` instead of `100%`, but this is definitely an improvement already. :+1: 

grafik

PR #236, opened 2024-10-28
feat(website): add logging

### LINKED ISSUE #203: Website: add logging
We need some proper logging capabilities on the website.

I'd propose to do the following:
* Use a logging framework? (Loculus uses https://github.com/winstonjs/winston AFAIK).
* Add a logging endpoint so that we can get the client side logs to our server.

fengelniederhammer:
resolves #203

### Summary

* Using https://github.com/winstonjs/winston
* For client side code, there is now `/api/log` that accepts POST requests with log messages.

### DIFF
--- a/website/.env.example
+++ b/website/.env.example
@@ -1,6 +1,8 @@
 CONFIG_DIR=../backend/src/main/resources/
 BACKEND_URL=http://localhost:8080
 DASHBOARDS_ENVIRONMENT=dashboards-dev
+LOG_DIR=./logs
+LOG_LEVEL=info
 
 GITHUB_CLIENT_ID="Check https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app"
 GITHUB_CLIENT_SECRET="put a secret of you GitHub app here"
--- a/website/src/logger.ts
+++ b/website/src/logger.ts
@@ -1,0 +1,45 @@
+import winston, { type Logger } from 'winston';
+import DailyRotateFile from 'winston-daily-rotate-file';
+
+import { type InstanceLogger } from './types/logMessage.ts';
+
+let _logger: Logger | undefined;
+
+const getLogger = (): Logger => {
+    if (_logger === undefined) {
+        const logPath = import.meta.env.LOG_DIR;
+        const serverLogFile = `${logPath}/dashboards_website.log`;
+        const transports: winston.transport[] = [new winston.transports.Console()];
+
+        if (logPath !== undefined) {
+            transports.push(
+                new DailyRotateFile({
+                    filename: serverLogFile,
+                    zippedArchive: true,
+                    datePattern: 'YYYY-MM-DD-HH-MM',
+                    maxFiles: 7,
+                }),
+            );
+        }
+
+        _logger = winston.createLogger({
+            level: import.meta.env.LOG_LEVEL ?? 'info',
+            format: winston.format.combine(winston.format.timestamp(), winston.format.json()),
+            transports,
+        });
+    }
+    return _logger;
+};
+
+/**
+ * This can only be used in server-side code.
+ * In client-side code, use `getClientLogger` instead.
+ */
+export const getInstanceLogger = (instance: string): InstanceLogger => {
+    return {
+        error: (message: string) => getLogger().error(message, { instance }),
+        warn: (message: string) => getLogger().warn(message, { instance }),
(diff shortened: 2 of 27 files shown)


website/src/logger.ts
```diff
@@ -0,0 +1,45 @@
...
+import DailyRotateFile from 'winston-daily-rotate-file';
+
+import { type InstanceLogger } from './types/logMessage.ts';
+
+let _logger: Logger | undefined;
+
+const getLogger = (): Logger => {
+    if (_logger === undefined) {
+        const logPath = import.meta.env.LOG_DIR;
+        const serverLogFile = `${logPath}/website.log`;
```
JonasKellerer, line 11:
```suggestion
        const serverLogFile = `${logPath}/dashboards_website.log`;
```
We have also other websites on the server. Just to make it more cleare where the log comes from.
Or we find a name for the backend and the website ;)

website/src/pages/api/log.ts
```diff
@@ -0,0 +1,26 @@
+import type { APIRoute } from 'astro';
+
+import { getInstanceLogger } from '../../logger.ts';
+import { type InstanceLogger, logMessageSchema } from '../../types/logMessage.ts';
+
+export const POST: APIRoute = async ({ request }) => {
+    const body = await request.json();
```
JonasKellerer, line 7: Currently, everybody can access this endpoint. Should we restrict the access to only logged in users? This would prevent us from getting error messages from the organism pages. But in theory one can flood our endpoint his way?

website/src/types/ProblemDetail.ts
```diff
@@ -2,11 +2,11 @@ import { z } from 'zod';
 
 export const problemDetailSchema = z
     .object({
-        type: z.string().url().optional(),
+        type: z.string().optional(),
         title: z.string().optional(),
         status: z.number().int().optional(),
         detail: z.string().optional(),
-        instance: z.string().url().optional(),
+        instance: z.string().optional(),
```
JonasKellerer, line 9: From the docs it seems that instance and the type should be an uri. Why did you change it here?

website/.env.example
```diff
@@ -1,6 +1,7 @@
 CONFIG_DIR=../backend/src/main/resources/
 BACKEND_URL=http://localhost:8080
 DASHBOARDS_ENVIRONMENT=dashboards-dev
+LOG_DIR=./logs
```
JonasKellerer, line 4: I think we should also add the log level here, since we have no other place to document it.

website/src/middleware/loggingMiddleware.ts
```diff
@@ -0,0 +1,13 @@
+import { defineMiddleware } from 'astro/middleware';
+
+import { getInstanceLogger } from '../logger.ts';
+
+const logger = getInstanceLogger('RequestLoggingMiddleware');
+
+export const loggingMiddleware = defineMiddleware(async (context, next) => {
+    const response = await next();
+    logger.info(
```
JonasKellerer, line 9: Do we really want to log every page load? I think it would be nice on debug.

JonasKellerer (commented):
Nice work.
I think the only thing we need to discuss, is whether we want to log every request to our page. And if everybody should be able to access our /api/log endpoint.

website/src/types/ProblemDetail.ts
fengelniederhammer: It wasn't a URL in fact. 

Also this should be as resilient as possible. Here (on the website), it doesn't matter what kind of string it is. But it should not fail to parse objects, because then our error handling fails (which is bad!).

website/src/pages/api/log.ts
fengelniederhammer: As discussed: 
* We want that every page can send log messages, also pages where users don't need to be logged in.
* There is no other proper protection, is it should be publicly available
* We only write log messages to the log.  We don't execute code or similar (assuming that winston doesn't have vulnerabilities). So we have a risk of being DDoSed / spamed, but otherwise we should be fine. DDoSing our server is quite easy anyway, because we have many LAPIS/SILO instances that do heavy computations.

We leave this endpoint open.

website/src/middleware/loggingMiddleware.ts
fengelniederhammer: As discussed: For now we'll leave it on `info`, until we notice that it's too much.

PR #265, opened 2024-11-07
feat(website): mark external navigation links

fengelniederhammer:
resolves #

### Summary

Using https://icon-sets.iconify.design/mdi/external-link/ to make it more obvious that you will leave the page when clicking the CovSpectrum link.

### Screenshot

grafik

grafik

### DIFF
--- a/website/src/layouts/base/header/MegaMenu.tsx
+++ b/website/src/layouts/base/header/MegaMenu.tsx
@@ -36,13 +36,18 @@ export function MegaMenuListEntry({
     href,
     label,
     className,
+    externalLink = false,
 }: WithClassName<{
     href: string;
     label: string;
+    externalLink?: boolean;
 }>) {
     return (
         <li>
-            <a className={`p-2 hover:underline hover:decoration-4 ${className}`} href={href}>
+            <a
+                className={`flex items-center p-2 hover:underline hover:decoration-4 ${className} ${externalLink ? 'after:iconify after:ml-0.5 after:mdi--external-link' : ''}`}
+                href={href}
+            >
                 {label}
             </a>
         </li>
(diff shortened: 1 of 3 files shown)


website/src/layouts/base/header/MegaMenu.tsx
```diff
@@ -36,13 +36,18 @@ export function MegaMenuListEntry({
...
 }: WithClassName<{
     href: string;
     label: string;
+    externalLink?: boolean;
 }>) {
     return (
         <li>
-            <a className={`p-2 hover:underline hover:decoration-4 ${className}`} href={href}>
+            <a
+                className={`p-2 hover:underline hover:decoration-4 ${className} ${externalLink ? 'after:iconify after:mdi--external-link' : ''}`}
```
JonasKellerer, line 48: Could you increase the spacing between the icon and the text? It sticks directly to it...
Maybe we need then a div and the anchor for this?

JonasKellerer: Do we want to align the text and the icon vertically?

website/src/pages/index.astro
```diff
@@ -79,7 +79,7 @@ import { ServerSide } from '../views/serverSideRouting';
                                     <div class='mr-2 w-4 border-b-2 border-b-black pt-[0.8rem]' />
                                     <a
                                         href='https://cov-spectrum.org'
-                                        class='hover:underline hover:decoration-lime-200 hover:decoration-4'
+                                        class={`hover:underline ${organism.menuListEntryDecoration} after:iconify after:mdi--external-link hover:decoration-4`}
```
JonasKellerer, line 82: Same as above.

JonasKellerer (changes requested):
Nice. Great idea.
I only have comments to the positioning :)

website/src/layouts/base/header/MegaMenu.tsx
fengelniederhammer: Actually I like that it's quite close to the link. I gave it a 0.5 margin now.

And I centered it.

JonasKellerer (approved):
I like it. :)

PR #269, opened 2024-11-09
234 show pathogen specific data source footer

### LINKED ISSUE #225: Improve footer design
Part of this ticket is to find out an "improved" design. 

But at least this should be done.
Currently, the footer floats to the center of the page, when there is not enough content on the page. It would be nice, that it stays on the bottom of the page.

### LINKED ISSUE #234: Show pathogen-specific data source footer
At the moement, we show in the footer "Enabled by data from [INSDC](https://www.insdc.org/)". As we now use data from Pathoplexus for WNV, I suggest

- show "Enabled by data from INSDC and Pathoplexus" on non-pathogen-specific pages (e.g., the landing page)
- show "Enabled by data from <the corresponding source>" on pathogen-specific pages

JonasKellerer:
resolves #234 
resolves #225

### Summary

This PR changes the design of the footer. There is now a primary footer, which should be shown on every page. It is vertically centered, and a secondary footer with additional information for each single page. For now this only holds the origin of the data. 

The footer is now positioned always at the bottom of the page.

### Screenshot

On home page

grafik

On organism page

grafik

On subscription page

grafik

### DIFF
--- a/website/src/layouts/base/footer/DataOriginLink.astro
+++ b/website/src/layouts/base/footer/DataOriginLink.astro
@@ -1,0 +1,11 @@
+---
+interface Props {
+    href: string;
+}
+
+const { href } = Astro.props;
+---
+
+<a class='underline after:iconify after:ml-0.5 after:mdi--external-link' href={href}>
+    <slot />
+</a>
--- a/website/src/layouts/base/footer/FooterNavigation.astro
+++ b/website/src/layouts/base/footer/FooterNavigation.astro
@@ -1,0 +1,10 @@
+---
+
+---
+
+<div class='flex min-h-32 items-center justify-center gap-6 p-2'>
+    <a href='/'>GenSpectrum</a>
+    <a class='flex items-center' href='https://github.com/GenSpectrum'>
+        <div class='iconify size-8 mdi--github'></div>
+    </a>
+</div>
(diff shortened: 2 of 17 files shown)


chaoran-chen:
Thanks! Looks great! One small request: could we write "INSDC and Pathoplexus", please? "INSDC, Pathoplexus" looks a bit like as INSDC is a division within Pathoplexus.

website/src/layouts/base/footer/DataOriginLink.astro
fengelniederhammer, line 1: We could also make this an `ExternalLink` and add the external link icon ;)

website/src/layouts/base/footer/FooterNavigation.astro
```diff
@@ -0,0 +1,12 @@
+---
+
+---
+
+<div class='flex min-h-32 items-center justify-center gap-2 p-2'>
+    <a href='/'>GenSpectrum</a>
+    <div>
```
fengelniederhammer, line 7: Do we need this div?

fengelniederhammer, line 7: I think a little marging between the two wouldn't hurt.

JonasKellerer: No :see_no_evil:

JonasKellerer:
It now looks like this:
grafik

PR #286, opened 2024-11-12
fix: config after changing the DNS entry for lapis.genspectrum.org from s1 to prod

fengelniederhammer:
### Summary

The RSV instances seem to have no lineage field yet... We will need to fix that in the deployment.


JonasKellerer (approved):
LGTM

PR #290, opened 2024-11-13
Orange/Grey layout with GS-favicon

anna-parker:
resolves #

### Summary

Same principle as https://github.com/GenSpectrum/dashboards/pull/289 but just with a different title.

### Screenshot

Current design: https://staging.genspectrum.org/

image
image

### TO DO:
- [x] Make title text smaller on small screens
- [x] Add link to pathogens in mobile-menu (current grid layout is too large and not easily addable as it is astro)
- [x] Add login link in mobile-menu

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -80,7 +80,7 @@ For development, it might be useful to use your local version of the
 [dashboard-components](https://github.com/GenSpectrum/dashboard-components):
 
 0. Stop the current `npm run dev` process.
-1. Install the local components
+1. Install the local components (run `npm run build` in the components directory)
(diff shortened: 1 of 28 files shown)


JonasKellerer:
Great. I like the design. Thank you.

I have some suggestions. Could we maybe:
  - reduce the font size of the headline on the "home" page also on normal screens. Weight heavy on the page.
  - Align the icons for the "About us" and "Open Source Tooling" section with the heading
  - Increase the top margin of the Github Link
  - put the "home" page in a container, so it does not cover the full screen on wide screens

PR #294, opened 2024-11-14
chore(website): use node 22

fengelniederhammer:


JonasKellerer (approved):
Thanks. Works also on my machine.

PR #301, opened 2024-11-15
feat(website): remove unused onBlur value from GsMutationFilter as removed in https://github.com/GenSpectrum/dashboard-components/pull/511

anna-parker:
resolves #

### Summary

In https://github.com/GenSpectrum/dashboard-components/pull/511 we removed the blur dispatch event from the mutation filter component, this PR reflects this change in genspectrum dashboards. 

The dispatch event was unused and was not accurate (e.g. it did not fire if a user moved outside of the component if they moved outside after deleting mutations)

### DIFF
--- a/website/src/components/genspectrum/GsMutationFilter.tsx
+++ b/website/src/components/genspectrum/GsMutationFilter.tsx
@@ -11,13 +11,11 @@ export type MutationFilter = {
 export function GsMutationFilter({
     initialValue,
     width,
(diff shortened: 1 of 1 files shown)


fengelniederhammer (approved):

fengelniederhammer:
Doesn't matter so much in this repo: `chore` or `refactor` would have been better than `feat` for this commit.

PR #302, opened 2024-11-15
feat(website): add clade filter and table to H5N1

### LINKED ISSUE #179: Add clades for H5N1
We can add clade labels using
1.  [LABEL](https://wonder.cdc.gov/amd/flu/label/) built by the CDC to assign avian flu clades

2. [nextclade H5Nx dataset](https://github.com/nextstrain/nextclade_data/blob/master/data/community/moncla-lab/iav-h5/ha/2.3.2.1/README.md) is available here (it includes the 2014 clade split: see https://onlinelibrary.wiley.com/doi/full/10.1111/irv.12324, seems to be the last official WHO clade assignment), it was shown to be close to the CDC tool LABEL and we already use nextclade for other clades so it should be easy to add.

fengelniederhammer:
resolves #179

### Screenshot

Adding the filter:
grafik

Adding the table to aggregated by clade:
grafik

### DIFF
--- a/website/src/pages/flu/h5n1/single-variant.astro
+++ b/website/src/pages/flu/h5n1/single-variant.astro
@@ -5,7 +5,7 @@ import { SingleVariantPageStateSelector } from '../../../components/pageStateSel
 import { getDashboardsConfig } from '../../../config';
 import SingleVariantOrganismPageLayout from '../../../layouts/OrganismPage/SingleVariantOrganismPageLayout.astro';
 import { chooseGranularityBasedOnDateRange } from '../../../util/chooseGranularityBasedOnDateRange';
-import { defaultTablePageSize, getLineageFilterConfigs } from '../../../views/View';
+import { defaultTablePageSize, getLineageFilterConfigs, getLineageFilterFields } from '../../../views/View';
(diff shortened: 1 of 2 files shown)


JonasKellerer (approved):
LGTM

For Mpox we have something like "None" when there is no clade. For H5N1 it seems there is no such thing but a blank. Is this something we want? I think this would need to be changed in the preprocessing.

anna-parker:
Thanks! I just tested it locally and the list looks very long and a bit hard to use as it isn't sorted: 
image
could we maybe sort the clades alphabetically?

fengelniederhammer:
> could we maybe sort the clades alphabetically?

That would be something we need to do in the components. It would probably be good to sort it, but also rather low prio (IMO)

fengelniederhammer:
> For Mpox we have something like "None" when there is no clade. For H5N1 it seems there is no such thing but a blank. Is this something we want? I think this would need to be changed in the preprocessing.

Yes, we would need to do that somewhere in our Loculus setup...

anna-parker:
I created an issue for sorting the output: https://github.com/GenSpectrum/dashboard-components/issues/514 do think that will make the UI more user friendly

PR #316, opened 2024-11-19
Pastels option

anna-parker:
resolves #

### Screenshot

image


fengelniederhammer:
We merged #317 instead

PR #317, opened 2024-11-19
Colorful layout

anna-parker:
resolves #

### Summary

This uses the color scheme used in dashboards

### Screenshot

image


website/package.json
```diff
@@ -19,10 +19,10 @@
     },
     "dependencies": {
         "@astrojs/node": "^8.3.4",
-        "@auth/core": "^0.37.4",
```
fengelniederhammer, line 22: Let's please not downgrade this... I'd rather accept some "lint ignore" comments.

anna-parker: I tried to add ignore statements to get the tests to pass for about 2hours and then gave up - the linter really wants to check the directory, potentially we could add more patches. If you have more experience with this we can maybe try together in a call.

anna-parker: Thanks so much for the fix!!!

website/src/layouts/base/BaseLayout.astro
```diff
@@ -20,11 +20,10 @@ const { title, forceLoggedOutState = false } = Astro.props;
 <html lang='en'>
     <head>
         <meta charset='utf-8' />
-        <link rel='icon' type='image/png' href='/favicon.png' />
+        <link rel='icon' type='image/png' href='/virus-outline-colorful.svg' />
         <meta name='viewport' content='width=device-width' />
         <meta name='generator' content={Astro.generator} />
         <title>{title}</title>
```
anna-parker, line 26: wupps merge conflict
```suggestion
        <title>{title}</title>
        <script is:inline defer data-domain='genspectrum.org' src='https://plausible.io/js/script.js'></script>
```

website/src/layouts/base/header/Brand.tsx
```diff
@@ -1,10 +1,29 @@
...
     );
 }
+
+type BrandNameProps = {
+    fontsize?: string;
+};
+
+export function BrandName({ fontsize = 'font-bold' }: BrandNameProps) {
+    return (
+        <a href='/' className='ml-2'>
```
fengelniederhammer, line 18: Should we remove the `a` here? 
* In `Brand`, we can simply put the `BrandName` inside the other `a` that is already there.
* On the `index.astro` landing page, IMO we don't need the link at all.

website/src/layouts/base/header/Header.astro
```diff
@@ -1,20 +1,77 @@
...
 const { forceLoggedOutState = false } = Astro.props;
+
+const session = await getSession(Astro.request);
+const showLoggedInState = !forceLoggedOutState && session?.user !== undefined;
 ---
 
 <header class='sticky top-0 z-10 border-b bg-white'>
-    <div class={`flex ${headerHeight} items-center justify-between`}>
-        <Brand name='GenSpectrum' />
+    <div class={`flex ${headerHeight} subtitle hidden md:flex md:z-6 gap-4 items-center justify-between`}>
```
fengelniederhammer, line 61: `flex` is overwritten by `hidden` anyway
```suggestion
    <div class={`${headerHeight} subtitle hidden md:flex md:z-6 gap-4 items-center justify-between`}>
```

fengelniederhammer, line 61: Is `subtitle` doing anything? I didn't see any effect in the browser
```suggestion
    <div class={`flex ${headerHeight} hidden md:flex md:z-6 gap-4 items-center justify-between`}>
```

fengelniederhammer, line 61: Why do we need `md:z-6`?

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+    'Compare side-by-side': 'mdi--compare-horizontal',
+    'Browse data': 'mdi--database-outline',
+};
+/* eslint-enable */
 ---
 
 <style>
+    a {
+        @apply text-slate-500 hover:underline;
+    }
```
fengelniederhammer, line 23: This is repeated from `data.astro`. Should we maybe extract it to a separate component instead (Astro or React) use that when we want to have links?

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+
+    .items a {
+        @apply text-black hover:underline hover:decoration-4;
+    }
```
fengelniederhammer, line 27: What is this part doing? If I remove it, then I don't see a difference in the browser (maybe I'm missing something though)

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+            @apply text-center text-6xl;
+        }
+
+        .largerLogo h2 {
+            @apply text-center text-2xl;
+        }
+
+        .infoSection h1 {
+            @apply text-2xl;
+        }
```
fengelniederhammer, line 50: I would prefer to inline those (instead of mixing inline CSS classes and this "global" stylesheet here)

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
 
-        <ul class='ml-6 mt-2 list-disc'>
-            <li>
-                …is a fully interactive platform aiming to help scientists and public health professionals query and
-                analyze pathogen sequences.
-            </li>
-            <li>…is open source, the code is available on <a href='https://github.com/GenSpectrum'>GitHub</a>.</li>
-            <li>
-                …uses data from the <a href='https://www.insdc.org/'
+    <div class='infoSection mx-auto my-12 flex max-w-[1200px] px-8'>
```
fengelniederhammer, line 65:
```suggestion
    <div class='infoSection mx-auto my-12 sm:flex max-w-[1200px] px-8'>
```
To get
grafik
instead of
grafik

(which is IMO easier to read on small screens)

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+        <div class='ml-8 basis-1/2 text-balance'>
+            <div class='items-left my-8 flex'>
```
fengelniederhammer, line 67: To align the icon with the text (vertically):
```suggestion
            <div class='items-center my-8 flex'>
```

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+                we are grateful for our many <a
+                    href='https://github.com/GenSpectrum/cov-spectrum-website?tab=readme-ov-file#contributors-'
+                >
+                    contributors
+                </a> from around the globe. We are funded by the <a href='https://ethz.ch/en.html'>ETH Zurich</a> and the
+                <a href='https://www.cdc.gov/'>CDC</a>.
+            </span>
+        </div>
+        <div class='ml-8 basis-1/2 text-balance'>
+            <div class='items-left my-8 flex'>
```
fengelniederhammer, line 83:
```suggestion
            <div class='items-center my-8 flex'>
```

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
-                analyze pathogen sequences.
-            </li>
-            <li>…is open source, the code is available on <a href='https://github.com/GenSpectrum'>GitHub</a>.</li>
-            <li>
-                …uses data from the <a href='https://www.insdc.org/'
+    <div class='infoSection mx-auto my-12 flex max-w-[1200px] px-8'>
+        <div class='ml-8 basis-1/2 text-balance'>
+            <div class='items-left my-8 flex'>
+                <div class='iconify size-7 mdi--information-outline'></div>
+                <h1 class='ml-2 text-3xl text-slate-600'>About Us</h1>
```
fengelniederhammer, line 69: Nitpicking, since the `<hN>` tags are for semantics only when using Tailwind: Isn't this more like a h3 rather than h1?

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+            </div>
+            <span
+                >An interactive platform to help scientists and public health professionals query and analyze pathogen
+                sequences. The core team is at the <a href='https://bsse.ethz.ch/cevo'>cEVO group </a> at ETH Zurich, but
+                we are grateful for our many <a
+                    href='https://github.com/GenSpectrum/cov-spectrum-website?tab=readme-ov-file#contributors-'
+                >
+                    contributors
+                </a> from around the globe. We are funded by the <a href='https://ethz.ch/en.html'>ETH Zurich</a> and the
+                <a href='https://www.cdc.gov/'>CDC</a>.
```
fengelniederhammer, line 79: The old text mentions CDC and Office of Advanced Molecular Detection separately - I'm not sure whether we should keep it that way?

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
+            </span>
+        </div>
+        <div class='ml-8 basis-1/2 text-balance'>
+            <div class='items-left my-8 flex'>
+                <div class='iconify size-7 mdi--wrench-outline'></div>
+                <h1 class='ml-2 text-3xl text-slate-600'>Open Source Tooling</h1>
+            </div>
+            <span
+                >Code is available on <a href='https://github.com/GenSpectrum'>GitHub</a>, with customizable components
```
fengelniederhammer, line 88: We changed the link in the footer, but apparently forgot the one in the text here - I think it would also be better to link the the repo instead of the project:
```suggestion
                >Code is available on <a href='https://github.com/GenSpectrum/dashboards'>GitHub</a>, with customizable components
```

website/src/pages/index.astro
```diff
@@ -1,88 +1,134 @@
...
-            </li>
-        </ul>
+            </span>
+        </div>
     </div>
 
-    <h1>Pathogens</h1>
-    <div class='mx-8 mb-8 flex flex-wrap justify-center gap-x-4 gap-y-8'>
+    <div class='mx-auto mb-8 mt-8 flex max-w-[1200px] items-center justify-center'>
+        <div class='iconify size-7 mdi--virus-outline'></div>
```
fengelniederhammer, line 100: Ah now I see :D You took this icon an colored it for our logo :+1:

website/src/types/Organism.ts
```diff
@@ -13,36 +13,46 @@ export const organismConfig = {
...
     [Organisms.rsvA]: {
         organism: Organisms.rsvA,
         pathFragment: 'rsv-a',
         label: 'RSV-A',
-        backgroundColor: 'bg-violet-200',
-        menuListEntryDecoration: 'hover:decoration-violet-200',
+        backgroundColor: 'bg-indigoMuted',
+        backgroundColorFocus: 'group-hover:bg-indigo',
+        menuListEntryDecoration: 'hover:decoration-indigo',
+        borderEntryDecoration: 'hover:order-indigo',
```
fengelniederhammer, line 46:
```suggestion
        borderEntryDecoration: 'hover:border-indigo',
```

website/src/components/covidAnalyzeSingleVariantView/CollectionsList.tsx
```diff
@@ -131,7 +131,7 @@ function CollectionVariantList({ collection, organismsConfig }: CollectionVarian
             {variants.map((variant, index) => (
                 <button
                     key={`${variant.name}_${variant.query}_${index}`}
-                    className='border bg-white px-4 py-2 hover:bg-amber-200'
+                    className='border bg-white px-4 py-2 hover:bg-cyan'
```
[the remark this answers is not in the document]
anna-parker: Ah it does use our color scheme actually. I added the color scheme to the tailwind config - so it uses the cyan that I defined in the config.

website/src/pages/index.astro
anna-parker: Yup - I took the svg and added the colors in adobe illustrator (the icons are also available for commercial use which is nice)

website/src/layouts/base/header/Header.astro
anna-parker: Also not sure why I added this

website/src/pages/index.astro
anna-parker: It keeps the color of the names of the links in the pathogen cards black and makes sure that their colorful underlines are thicker

website/src/pages/index.astro
anna-parker: True! It is actually the same size and color as the sub-title (h2) so I will set them all as h2 - will also make the code a bit DRYer

website/src/pages/index.astro
anna-parker: I could place this in the `tailwind.css` file as a global default? Would that be ok?

website/src/pages/index.astro
anna-parker: done :-)

website/tailwind.config.mjs
```diff
@@ -1,11 +1,40 @@
 import { addIconSelectors } from '@iconify/tailwind';
 import daisyui from 'daisyui';
 
+const colors = {
+    indigo: '#473894', // rgba(51, 34, 136, 0.9)
+    green: '#117733',
+    cyan: '#35a6de', // I modified '#88CCEE' manually to make it darker
+    teal: '#44AA99',
+    olive: '#999933',
+    sand: '#DDCC77',
```
fengelniederhammer, line 10: I played around with this a bit and found `sand: '#BBAA55'` slightly better, because it increases the contrast:

Old:
grafik

New:
grafik

JonasKellerer (approved):
I like the new colors. 

Later on, we might use the tailwind functionality for muting colors.

PR #319, opened 2024-11-19
feat(website): delete mpox

fengelniederhammer:
### Summary

We'll bring it back once Pathplexus has launched Mpox and use its LAPIS instance. It was easiest to delete the whole thing. Other options would have been:
* Filtering out mpox in the navigation (seems fragile - easy to forget when bringing it back)
* deleting only the organism config (Typescript and the linter would have complained in various places)


JonasKellerer (approved):
LGTM

I think there is still two entries left on the "Data sources" page :)

fengelniederhammer:
> I think there is still two entries left on the "Data sources" page :)

Thanks, good catch :+1:

PR #322, opened 2024-11-20
Colorful layout linter fix

fengelniederhammer:
resolves #


website/patches/auth-astro+4.1.3.patch
anna-parker, line 1: Omg this is amazing!!! Can you show me how you did this later in a call?

fengelniederhammer: It's actually simple:
* edit `website/node_modules/auth-astro/client.ts` manually
* run `npx patch-package auth-astro`

anna-parker: And the manual import part? Did you just migrate to the function and paste it into the patch file or was it more involved?

website/src/types/Organism.ts
```diff
@@ -62,7 +62,7 @@ export const organismConfig = {
         backgroundColor: 'bg-indigoMuted',
         backgroundColorFocus: 'group-hover:bg-indigo',
         menuListEntryDecoration: 'hover:decoration-indigo',
-        borderEntryDecoration: 'hover:border-indigo',
+        borderEntryDecoration: 'hover:order-indigo',
```
anna-parker, line 65:
```suggestion
        borderEntryDecoration: 'hover:border-indigo',
```

PR #323, opened 2024-11-20
chore(website): patch auth-astro - manually import types from next-auth/react

fengelniederhammer:
### Summary

A copy of #322 to be merged into main instead of into another PR.

The dependency is not installed transitively - we don't want to install it ourselves We have to satisfy Typescript and the linter here. I originally fixed the build with auth-astro in #307, but this seems a better fix, since it simply removes the imports from the missing dependency.


anna-parker (approved):
Amazing thanks!

PR #324, opened 2024-11-20
feat(website): add a hamburger menu with links to all pages and login link

anna-parker:
resolves #

### DIFF
--- a/website/src/layouts/base/header/HamburgerMenu.tsx
+++ b/website/src/layouts/base/header/HamburgerMenu.tsx
@@ -1,0 +1,213 @@
+import type { Session } from '@auth/core/types';
+import { signIn } from 'auth-astro/client';
+import { useEffect, useState } from 'react';
+
+import { Hamburger } from './Hamburger';
+import { OffCanvasOverlay, useOffCanvas } from './OffCanvasOverlay';
+
+type PathogenMegaMenuSection = {
+    headline: string;
+    headlineBackgroundColor: string;
+    navigationEntries: PathogenMegaMenuItems[];
+};
+
+export type PathogenMegaMenuItems = {
+    label: string;
+    href: string;
+    underlineColor: string;
+    externalLink: boolean;
+};
+
+type HamburgerMenuProps = {
+    session: Session | null;
+    pathogenMegaMenuSections: PathogenMegaMenuSection[];
+    showLoggedInState: boolean;
+};
+
+function HamburgerMenu({ session, pathogenMegaMenuSections, showLoggedInState }: HamburgerMenuProps) {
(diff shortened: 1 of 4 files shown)


website/src/layouts/base/header/HamburgerMenu.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+function OffCanvasNavSection({ section }: OffCanvasNavSectionProps) {
+    const [viewDiv, setViewDiv] = useState(false);
+    const reverseView = () => {
+        setViewDiv(!viewDiv);
+    };
+
+    return (
+        <div key={section.headline}>
+            <button onClick={reverseView} className={`ml-8 py-3 ${viewDiv === true ? 'font-bold' : ''}`}>
```
fengelniederhammer, line 146: I think it would be good if those items had some sort of arrow icon that indicate that they open another menu (instead of directly linking somewhere)

website/src/layouts/base/header/HamburgerMenu.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+type OffCanvasNavSectionProps = {
+    section: PathogenMegaMenuSection;
+};
+
+function OffCanvasNavSection({ section }: OffCanvasNavSectionProps) {
+    const [viewDiv, setViewDiv] = useState(false);
+    const reverseView = () => {
+        setViewDiv(!viewDiv);
+    };
```
fengelniederhammer, line 142: Instead of using JS for toggling this menu, we could a CSS-only solution, which is probably simpler. Daisy UI's collapse items are usually good for those things:
https://daisyui.com/components/collapse/

website/src/layouts/base/header/OffCanvasOverlay.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export function OffCanvasOverlay({ onClick, className }: Props): JSX.Element {
+    return <div className={`fixed inset-0 z-20 bg-gray-800 bg-opacity-50 ${className ?? ''}`} onClick={onClick} />;
+}
+
+export const useOffCanvas = (initialIsOpen = false) => {
+    const [isOpen, setOpen] = useState(initialIsOpen);
+
+    const open = useCallback(() => {
+        document.body.style.overflow = 'hidden'; // This makes the background not scrollable
```
fengelniederhammer, line 16: How important is that feature actually? It seems overly complicated to me just for the sake of making the background not scrollable.

An alternative solution would be to make the open hamburger menu use the full page, then the problem whether the background is scrollable or not simply disappears.

fengelniederhammer (changes requested):

fengelniederhammer:
We'll merge #327 instead

PR #325, opened 2024-11-20
fix(backend): bring back mpox config to fix backend startup

fengelniederhammer:
### Summary

We didn't notice in #319 that it breaks the backend.


JonasKellerer (approved):
LGTM

PR #327, opened 2024-11-20
feat(website): reactive hamburger menu

### LINKED ISSUE #220: Responsive header
Currently, the header is not responsive, meaning, on small (mobile) screens it is not very useable. Make it, so the navigation and the call to action is moved inside a hamburger menu to the right. There each menu entry should be inside a nested list. The login should be on the bottom.

anna-parker:
resolves #220 

### Summary

This basically reimplements the hamburger without react

### Missing items
1. Github icon is at the bottom of the page and fixing its position makes it visible even when hamburger is closed: 
image
2. Login should call previously implemented LoginButton but this is not working.


fengelniederhammer:
I changed it to full screen:

grafik

website/src/layouts/base/header/HamburgerMenu.astro
```diff
@@ -0,0 +1,73 @@
...
+
+const { forceLoggedOutState } = Astro.props;
+
+const pathogenMegaMenuSections = getPathogenMegaMenuSections();
+
+const session = await getSession(Astro.request);
+const showLoggedInState = !forceLoggedOutState && session?.user !== undefined;
+---
+
+<div>
```
JonasKellerer, line 21:
```suggestion
<nav>
```
So screen reader can see that this is a website navigation. (Has no other effect).

website/src/layouts/base/header/Navigation.tsx
```diff
@@ -1,38 +1,12 @@
...
-        return {
-            headline: organism.label,
-            headlineBackgroundColor: organism.backgroundColor,
-            navigationEntries: megaMenuSections,
-        };
-    });
+    const pathogenMegaMenuSections = getPathogenMegaMenuSections();
 
     return (
         <nav className=''>
```
JonasKellerer, line 12:
```suggestion
        <nav>
```

website/src/layouts/base/header/Header.astro
```diff
@@ -1,20 +1,27 @@
...
-
 const { forceLoggedOutState = false } = Astro.props;
+
+const zIndexHigherThanMutationFilterInfoButton = 'z-20';
 ---
 
-<header class='sticky top-0 z-10 border-b bg-white'>
-    <div class={`flex ${headerHeight} items-center justify-between`}>
+<header class={`sticky top-0 ${zIndexHigherThanMutationFilterInfoButton} border-b bg-white`}>
+    <div class={`${headerHeight} hidden sm:flex md:z-6 gap-4 items-center justify-between`}>
```
JonasKellerer, line 17:
```suggestion
    <div class={`${headerHeight} hidden sm:flex gap-4 items-center justify-between`}>
```
Is the `md:z-6` necessary?

website/src/layouts/base/header/HamburgerMenuItem.astro
```diff
@@ -0,0 +1,63 @@
...
+          }
+        | {
+              type: 'section';
+              navigationEntries: {
+                  label: string;
+                  href?: string;
+                  externalLink?: boolean;
+              }[];
+          };
+    style: 'toplevel' | 'subitem';
```
JonasKellerer, line 17: I could not find, where `subitem` is used.

website/src/layouts/base/header/HamburgerMenu.astro
```diff
@@ -0,0 +1,73 @@
...
+        <HamburgerMenuItem content={{ type: 'item', href: Page.dataSources }} style='toplevel'
+            >Data Sources
+        </HamburgerMenuItem>
+        {
+            showLoggedInState ? (
+                <HamburgerMenuItem
+                    content={{
+                        type: 'section',
+                        navigationEntries: [
+                            { label: session.user!.name! },
```
JonasKellerer, line 54: It is a bit strange, that the username has the same style as the others links. But I think this is fine for now, since we do not want to launch with the login anyway.

website/src/layouts/base/header/HamburgerMenuItem.astro
```diff
@@ -0,0 +1,63 @@
...
+                        content.externalLink === true
+                            ? 'after:iconify after:ml-0.5 after:align-text-bottom after:mdi--external-link'
+                            : ''
+                    }
+                    href={content.href}
+                >
+                    <slot />
+                </a>
+            )
+        ) : (
```
JonasKellerer, line 45: It would be easier, to follow, when this is its own astro file. Then we have "HamburgerMenuItem" and "HamburgerMenuSection". They don't share so much here.

website/src/layouts/base/header/HamburgerMenuItem.astro
```diff
@@ -0,0 +1,63 @@
...
+            <details class='collapse collapse-arrow m-0 p-0'>
```
JonasKellerer, line 46: Has the "m-0" and "p-0" an effect? It seems only for the summary below,

JonasKellerer (approved):
Looks great. :)

Only some minor refactors and cleanups leftover.

website/src/layouts/base/header/Header.astro
fengelniederhammer: No, good catch :)

website/src/layouts/base/header/HamburgerMenu.astro
fengelniederhammer: Let's improve that later :+1:

PR #353, opened 2024-11-26
fix(website): fix reference error in built artifact

fengelniederhammer:
resolves #

### Summary

The built website would crash with "ReferenceError: can't access lexical declaration '$e' before initialization" (in Firefox) $e was one of the abstract view classes.
I guess this was due to a cyclic dependency (view requiring state handlers requiring view), but I'm not 100% sure. It works with the abstract view classes extracted to a separate file though.


JonasKellerer (approved):
LGTM. Thank you.

PR #354, opened 2024-11-26
feat(website): show more useful error messages to the user

### LINKED ISSUE #256: Provide more useful error messages to users
Currently, we only show something like "Ooops, error" to users, when something unexpected happens on the website.

Show something more useful to users:
- When the backend is down (show "backend down")
- when the backend returns an error (show the error detail)
- when forseeable errors happen on the website (tell them what went wrong, because we know what went wrong)
  - This includes all places where the website itself `throw`s.
- else fall back to "something unexpected happened, and we won't tell you what".

We could solve this by having a class `UserFacingError extends Error`. Whenever we catch a `UserFacingError`, we can display it's error message, otherwise we display "expected error occured".
Then we have quite good control over whether to display the error to the user or not. Also by default JS related `TypeError`s won't be displayed.

fengelniederhammer:
resolves #256

### Screenshot

On client side (where the error can be seen in console and network tab anyway): Now shows the error message 
* in the error information that's supposed to be copied to our GitHub issues
* in the error toast, if it should be user facing
 
grafik

On server side: Shows the error message
* in the error information,  if it should be user facing
* in the error toast, if it should be user facing
grafik

### DIFF
--- a/website/src/views/PageStateHandler.ts
+++ b/website/src/views/PageStateHandler.ts
@@ -23,7 +23,8 @@ import {
     setSearchFromLocation,
 } from './helpers.ts';
 import { compareVariantsViewKey } from './routing.ts';
-import { Organisms } from '../types/Organism.ts';
+import { UserFacingError } from '../components/ErrorReportInstruction.tsx';
+import { organismConfig } from '../types/Organism.ts';
 
 export interface PageStateHandler<PageState extends object> {
     parsePageStateFromUrl(url: URL): PageState;
@@ -143,6 +144,8 @@ function toLapisFilterWithoutVariant(
     };
 }
 
+const $ = '$';
+
 export abstract class CompareVariantsStateHandler<ColumnData extends BaselineAndVariantData = BaselineAndVariantData>
     implements PageStateHandler<CompareVariantsData<ColumnData>>
 {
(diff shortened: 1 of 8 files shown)


website/src/views/PageStateHandler.ts
```diff
@@ -143,6 +144,8 @@ function toLapisFilterWithoutVariant(
     };
 }
 
+const $ = '$';
```
JonasKellerer, line 147: I don't understand this construction. Is this the delimiter?

JonasKellerer (approved):
LGTM

website/src/views/PageStateHandler.ts
fengelniederhammer: Yes, this is the delimiter. I couldn't find a better name, so I took IntelliJ's suggestion. Apparently `$` is a valid variable name in JS.

JonasKellerer: Can we not name it delimiter?

fengelniederhammer: I wanted to be a bit more specific (which delimiter?). If we find a good name, sure. In the meanwhile, I don't find `$` so bad.

JonasKellerer: How about `filterSearchParamDelimiter`?

PR #356, opened 2024-11-27
346 cant add column in empty compare side by side view

### LINKED ISSUE #346: Can't add column in empty compare side by side view
* Go to covid "Compare variants side-by-side"
* click on breadcrumb:
grafik
* now you are on an essentially empty page.
* click "add column"
* you are still on an empty page and the url looks broken: `/covid/compare-side-by-side?region%24-Infinity=Europe&date%24-Infinity=Last+year`

grafik

JonasKellerer:
resolves #346

### Summary
Now, when clicking on the compare variants view on the breadcrumbs, it directs to the default page. 

I also added a fix: when users types the url without the filter: `/flu/h5n1/compare-side-by-side` it then shows an empty compare view and the user can click on add, and gets one entry with the content of the `epmtyColumnData`.


website/src/views/PageStateHandler.ts
```diff
@@ -209,9 +209,14 @@ export abstract class CompareVariantsStateHandler<ColumnData extends BaselineAnd
     }
 
     public addEmptyFilter(pageState: CompareVariantsData<ColumnData>): CovidCompareVariantsData {
-        const lastId = Math.max(...Array.from(pageState.filters.keys()));
+        const newId = () => {
+            if (pageState.filters.size === 0) {
+                return 0;
+            }
+            return Math.max(...Array.from(pageState.filters.keys())) + 1;
+        };
```
fengelniederhammer, line 217:
```suggestion
        const newId = pageState.filters.size === 0 ? 0 : Math.max(...Array.from(pageState.filters.keys())) + 1;
```

website/src/views/PageStateHandler.ts
```diff
@@ -209,9 +209,14 @@ export abstract class CompareVariantsStateHandler<ColumnData extends BaselineAnd
...
 
-        return this.setFilter(pageState, this.getEmptyColumnData(), lastId + 1);
+        return this.setFilter(pageState, this.getEmptyColumnData(), newId());
```
fengelniederhammer, line 219:
```suggestion
        return this.setFilter(pageState, this.getEmptyColumnData(), newId);
```

fengelniederhammer (approved):

website/src/views/PageStateHandler.ts
JonasKellerer: I had it like this and then moved to the function to make it clearer. But since you also prefer this, I changed it back. :)

Issue #352, opened 2024-11-26
Compare to baseline: implement spike


fengelniederhammer:
Implement a design proposal. Maybe something like:
* have a search bar on the left side. Similar to single variant page, with a "add variant" button at the bottom. With "baseline variant filter" at the top.
* Show prevalence over time

Issue #351, opened 2024-11-26
Compare variants view: implement spike


fengelniederhammer:
Implement a design proposal. Maybe something like:
* have a search bar on the left side. Similar to single variant page, with a "add variant" button at the bottom
* Show prevalence over time
* Show mutation comparison

PR #357, opened 2024-11-27
350 single variant page dont show plots when no variant selected

### LINKED ISSUE #350: Single variant page: don't show plots when no variant selected
When no variant is selected, then don't show the plots on the single variant page. Instead show a text that one should add something.

JonasKellerer:
resolves #350

### Summary

There is now a display on the single variant view, which shows, when no variant is selected. When there is a variant selected, it shows the data.

### Screenshot

grafik


JonasKellerer:
@chaoran-chen I find it hard to write a nice message to the users. What do you think about the current version?

JonasKellerer:
@fengelniederhammer Currently, I have only refactored the single variant view. I imagine there is more possible. I think we can find something better here.

website/src/components/views/analyzeSingleVariant/SelectVariant.astro
```diff
@@ -0,0 +1,14 @@
+---
+import { PageHeadline } from '../../../styles/containers/PageHeadline';
+---
+
+<div class='mt-20 flex min-h-64 flex-col items-center'>
+    <PageHeadline>Analyze a single variant</PageHeadline>
+    <div class='max-w-xl'>
+        <p class='mb-2'>To proceed, please select a variant filter.</p>
+        <p>
+            Note that the dataset filter is applied to the entire dataset. Adjust the dataset filter as needed to refine
```
fengelniederhammer, line 10: grafik

In the filter panel on the left, it's called "Filter dataset", but in the text we refer to it as "dataset filter". IMO we should make that consistent. I'd prefer changing it in the filter panel on the left, since "Dataset Filter" would match "Variant Filter" below better.

website/src/components/views/analyzeSingleVariant/SelectVariant.astro
```diff
@@ -0,0 +1,14 @@
+---
+import { PageHeadline } from '../../../styles/containers/PageHeadline';
+---
+
+<div class='mt-20 flex min-h-64 flex-col items-center'>
+    <PageHeadline>Analyze a single variant</PageHeadline>
+    <div class='max-w-xl'>
```
fengelniederhammer, line 7: I think this could use a little mx or px, so that the text has a little distance to the edge on small screens:
grafik

website/src/util/hasOnlyUndefinedValues.ts
```diff
@@ -0,0 +1,19 @@
+export function hasOnlyUndefinedValues(obj: any): boolean {
+    if (typeof obj !== 'object' || obj === null) {
+        return false;
+    }
```
fengelniederhammer, line 4:
```suggestion
export function hasOnlyUndefinedValues(obj: object): boolean {
    if (obj === null) {
        return true;
    }
```

website/src/util/hasOnlyUndefinedValues.ts
```diff
@@ -0,0 +1,19 @@
...
+            if (typeof value === 'object' && value !== null) {
+                if (!hasOnlyUndefinedValues(value)) {
+                    return false;
+                }
+            } else if (value !== undefined) {
+                return false;
+            }
+        }
+    }
+    return true;
```
fengelniederhammer, line 18: ChatGPT suggested:
```suggestion
    return Object.entries(obj).every(([_, value]) => {
        if (typeof value === 'object' && value !== null) {
            return hasOnlyUndefinedValues(value);
        }
        return value === undefined;
    });
```

website/src/util/hasOnlyUndefinedValues.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { describe } from 'vitest';
+
+import { hasOnlyUndefinedValues } from './hasOnlyUndefinedValues.ts';
+
+describe('hasOnlyUndefinedValues', (it) => {
```
fengelniederhammer, line 5: I think I would add two test cases:
* `const obj = null`
* `const obj = { a: null }`

website/src/util/hasOnlyUndefinedValues.ts
fengelniederhammer: Maybe this is even better:
```suggestion
export function hasOnlyUndefinedValues(obj: {}): boolean {
```

eslint will complain:
>The `{}` ("empty object") type allows any non-nullish value, including literals like `0` and `""`.
>- If that's what you want, disable this lint rule with an inline comment or configure the 'allowObjectTypes' rule option."

But I think in this case, it's actually what we want? It doesn't allow `null` and we don't care about values or keys, because we check them anyway.

website/src/pages/covid/single-variant.astro
```diff
@@ -66,110 +70,124 @@ const lineageFilterConfigs = getLineageFilterConfigs(
...
-                    lapisFilter={JSON.stringify(variantFilter)}
-                    sequenceType='amino acid'
-                    views='["grid", "table", "insertions"]'
-                    pageSize={defaultTablePageSize}
-                    width='100%'
-                    height='100%'></gs-mutations>
-            </ComponentWrapper>
+    {noVariantSelected && <SelectVariant slot='dataDisplay' />}
+    {
+        !noVariantSelected && (
```
fengelniederhammer, line 76:
```suggestion
    {noVariantSelected ? <SelectVariant slot='dataDisplay' />
    
        : (
```

website/src/components/views/analyzeSingleVariant/SelectVariant.astro
JonasKellerer: I changed it to "Dataset Filter", which we will also need in #352 and #351

website/src/util/hasOnlyUndefinedValues.ts
JonasKellerer: I took the version with `obj: object`, since `0` and `""` would be possible in the `{}` solution.

website/src/pages/covid/single-variant.astro
```diff
@@ -83,6 +83,7 @@ const lineageFilterConfigs = getLineageFilterConfigs(
                         lapisFilter: variantFilter,
                     })}
                     denominatorFilter={JSON.stringify(baselineFilter)}
+                    lapisDateField={view.organismConstants.mainDateField}
```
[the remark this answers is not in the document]
JonasKellerer: There is now https://github.com/GenSpectrum/dashboard-components/issues/554 which keeps track of validating the input.

fengelniederhammer (approved):

PR #359, opened 2024-11-28
351 compare variants view implement spike

JonasKellerer:
### Summary
This is a draft of the compare variants view to show the general design.

The general ui stays the same from the analyze single variant view, only the filter section is changed. There users can click to add more variants.

### Screenshot
grafik

### DIFF
--- a/website/src/components/ComponentsGrid.astro
+++ b/website/src/components/ComponentsGrid.astro
@@ -2,6 +2,6 @@
 
 ---
 
-<div class='grid grid-cols-[repeat(auto-fit,minmax(min(calc(100vw-16px),650px),1fr))] gap-x-4 gap-y-6'>
+<div class='grid grid-cols-[repeat(auto-fit,minmax(min(calc(100vw-16px),650px),1fr))] content-start gap-x-4 gap-y-6'>
     <slot />
 </div>
--- a/website/src/components/pageStateSelectors/VariantsSelector.tsx
+++ b/website/src/components/pageStateSelectors/VariantsSelector.tsx
@@ -1,0 +1,105 @@
+import { type VariantFilterConfig } from './CompareVariantsPageStateSelector.tsx';
+import { LineageFilterInput } from './LineageFilterInput.tsx';
+import { SelectorHeadline } from './SelectorHeadline.tsx';
+import type { Id } from '../../views/View.ts';
+import { getMutationFilter } from '../../views/helpers.ts';
+import { GsMutationFilter, type MutationFilter } from '../genspectrum/GsMutationFilter.tsx';
+
+function copyVariantFilterConfig(variantFilterConfig: VariantFilterConfig): VariantFilterConfig {
+    return {
+        mutationFilterConfig: {
+            ...variantFilterConfig.mutationFilterConfig,
+        },
+        lineageFilterConfigs: variantFilterConfig.lineageFilterConfigs.map((lineageFilterConfig) => {
+            return {
+                ...lineageFilterConfig,
+            };
+        }),
+    };
+}
+
+export function VariantsSelector({
+    variantFilterConfigs,
+    setVariantFilterConfigs,
+    emptyVariantFilterConfig,
+}: {
+    variantFilterConfigs: Map<Id, VariantFilterConfig>;
+    setVariantFilterConfigs: (variants: Map<Id, VariantFilterConfig>) => void;
+    emptyVariantFilterConfig: VariantFilterConfig;
+}) {
+    const removeVariant = (id: Id) => {
+        setVariantFilterConfigs(new Map(Array.from(variantFilterConfigs).filter(([key]) => key !== id)));
+    };
+
+    const addVariant = () => {
+        const newVariantFilterConfigs = new Map(variantFilterConfigs);
+
+        const newId = variantFilterConfigs.size === 0 ? 0 : Math.max(...Array.from(variantFilterConfigs.keys())) + 1;
+
+        newVariantFilterConfigs.set(newId, copyVariantFilterConfig(emptyVariantFilterConfig));
+
+        setVariantFilterConfigs(newVariantFilterConfigs);
+    };
+
+    const updateLineageFilterConfig = (id: Id, index: number, lineage: string | undefined) => {
+        const newVariantFilterConfigs = new Map(variantFilterConfigs);
+
+        const variantFilterConfig = newVariantFilterConfigs.get(id)!;
+
+        variantFilterConfig.lineageFilterConfigs[index].initialValue = lineage;
+
+        newVariantFilterConfigs.set(id, variantFilterConfig);
+
+        setVariantFilterConfigs(newVariantFilterConfigs);
+    };
+
+    const updateMutationFilter = (id: Id, mutation: MutationFilter | undefined) => {
+        const newVariantFilterConfigs = new Map(variantFilterConfigs);
+
+        const variantFilterConfig = newVariantFilterConfigs.get(id)!;
+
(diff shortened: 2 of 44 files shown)


JonasKellerer:
Thank you for the feedback.

Concerning the number format on the venn view: the number is indeed two thousand something. Only the display uses the locale delimiter. I created https://github.com/GenSpectrum/dashboard-components/issues/552 to fix this.

website/src/components/ComponentsGrid.astro
```diff
@@ -2,6 +2,6 @@
 
 ---
 
-<div class='grid grid-cols-[repeat(auto-fit,minmax(min(calc(100vw-16px),650px),1fr))] gap-x-4 gap-y-6'>
+<div class='grid grid-cols-[repeat(auto-fit,minmax(min(calc(100vw-16px),650px),1fr))] content-start gap-x-4 gap-y-6'>
```
JonasKellerer, line 5: I changed this, because the filter now increases the page size. The grid then tries to fill the elements evenly on the page. For only 3 graphs, this then no longer looks nice.

JonasKellerer:
I now finished the implementation. The view is now available for all pathogens.

Leftover points:
  - What name do we want to display next to the lineages? Currently, I used only use numbering.

JonasKellerer:
> We could also use it to label variants. It should be relatively easy to compute that programmatically from the filters?

Good suggestion. I also thought about it, but this does not work well with lineages from west-nile. Some of them are just numbers. Also, I wonder what we will do when we have clades and lineages mixed.

fengelniederhammer:
> > We could also use it to label variants. It should be relatively easy to compute that programmatically from the filters?
> 
> Good suggestion. I also thought about it, but this does not work well with lineages from west-nile. Some of them are just numbers. Also, I wonder what we will do when we have clades and lineages mixed.

Maybe we can make it view specific? For Covid, only the pango lineage should be good enough. In other cases we could use a similar format to what we have for Covid variant queries in LAPIS: `clade:5 + A123T` or `clade:5 + lineage:A.1`. I imagine that it could work.

website/src/views/View.ts
```diff
@@ -33,6 +33,13 @@ export type CompareSideBySideData<ColumnData extends BaselineAndVariantData = Ba
     filters: Map<Id, ColumnData>;
 };
 
+export type CompareVariantsData = {
+    variants: Map<Id, VariantFilter>;
+} & BaselineData;
+
+export type CovidCompareVariantsFilter = BaselineData & VariantData<LapisCovidVariantFilter>;
+export type CovidCompareVariantsData = CompareSideBySideData<CovidCompareVariantsFilter>;
```
fengelniederhammer, line 41: `CovidCompareVariantsData`  and `CovidCompareVariantsFilter` seems unused?

website/src/views/covid.ts
```diff
@@ -325,3 +325,9 @@ class CovidSequencingEffortsStateHandler
         return `${this.pathname}?${search}`;
     }
 }
+
+export class CovidCompareVariantsView extends GenericCompareVariantsView<CovidConstants> {
+    constructor(organismsConfig: OrganismsConfig) {
+        super(new CovidConstants(organismsConfig));
+    }
+}
```
fengelniederhammer, line 333: I'm not sure whether we directly want to support `variantQuery`s  here. But let's leave it as it is until we really need it. (I think that's the reason why we have separate logic in the other views for covid)

website/src/views/helpers.ts
```diff
@@ -114,6 +115,15 @@ export type LapisMutationQuery = {
     aminoAcidInsertions?: string[];
 };
 
+export function getMutationFilter(mutationFilter: LapisMutationQuery): MutationFilter {
```
fengelniederhammer, line 118: This is essentially a duplicate of `getLapisMutations`, but with a better return type. We don't need both functions, do we?

fengelniederhammer (approved):
Looks great :)

website/src/components/pageStateSelectors/VariantsSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+                                    key={lineageFilterConfig.lapisField}
+                                />
+                            ))}
+                            <GsMutationFilter
+                                initialValue={getMutationFilter(filterConfig.mutationFilterConfig)}
+                                onMutationChange={(mutation) => updateMutationFilter(id, mutation)}
+                            />
+                        </div>
+                    </div>
+                ))}
```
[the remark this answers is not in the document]
JonasKellerer: I would add a similar description as in #350. Then we don't show anything before the user has entered something. I would leave it like this, because the compare-side-by-side page works similarly. I added a ticket https://github.com/GenSpectrum/dashboards/issues/376

website/src/views/covid.ts
JonasKellerer: We have the variant query for covid, because we use it for the collections on the analyze-single-variant-view. But we have no input element to get the variant query otherwise. I think we leave it like this and only add it, when we have a variantQuery filter element on the page.

website/src/components/rsv/RsvSingleVariantPage.astro
fengelniederhammer, line 1: Let's delete this zombie (probably got resurrected during rebase)

website/src/views/PageStateHandler.ts
```diff
@@ -305,3 +282,162 @@ export class GenericCompareVariantsStateHandler extends CompareVariantsStateHand
...
+    }
+}
+
+function decodeFiltersFromSearch(search: URLSearchParams, handleNonMatchingKey: (key: string) => void = () => {}) {
+    const filterMap = new Map<Id, Map<string, string>>();
+
+    for (const [key, value] of search) {
+        const keySplit = key.split($);
+        if (keySplit.length !== 2) {
+            handleNonMatchingKey(key);
```
fengelniederhammer, line 414: Do we actually need that difference? Or could we throw in any case (or ignore the error)? (I think I'd prefer throwing, but not sure)

website/src/views/PageStateHandler.ts
```diff
@@ -305,3 +282,162 @@ export class GenericCompareVariantsStateHandler extends CompareVariantsStateHand
...
+    const search = new URLSearchParams();
+    for (const [id, filter] of filters) {
+        for (const [key, value] of filter) {
+            search.append(`${key}${$}${id}`, value);
+        }
+    }
+    return search;
+}
+
+type NamedLapisFilter = {
```
fengelniederhammer, line 440: We should export that from the components, too (not a task for now, but to keep it in mind)

fengelniederhammer (approved):
grafik
The label can get cut off when it's too long. But we can fix that later.

website/src/components/rsv/RsvSingleVariantPage.astro
JonasKellerer: ... not again ... sry.

website/src/views/PageStateHandler.ts
JonasKellerer: I added https://github.com/GenSpectrum/dashboard-components/issues/556 to keep track of it.

JonasKellerer: We had already a ticket for that: https://github.com/GenSpectrum/dashboard-components/issues/248. I've closed the new one.

website/src/views/PageStateHandler.ts
JonasKellerer: I now removed the error.

fengelniederhammer: Ah I see. On the compare side-by-side page, all keys had to be split. I'm not sure whether it's better to silently ignore unknown parameters or throwing an error. Let's see...

PR #364, opened 2024-11-28
feat(website): add statistics to single variant views

### LINKED ISSUE #336: Show total number of sequences
For all the views that we currently have, it would be nice to show the total number of filtered sequences. At the moment, we only see the number of sequences stratified by time / location / etc. but can't see the total number.

(Question: are there other useful metrics to show?)

JonasKellerer:
resolves #336

### Summary

Adds the statistics component to all "analyze single variant" pages.

Also ensures that the loading window is as large as the component and fixes a minor issue, that the padding of the "mutations-over-time" component was different from the rest of the grid.

### Screenshot

On large screen:
grafik

On small screen:
grafik


website/package.json
```diff
@@ -21,7 +21,7 @@
     "dependencies": {
         "@astrojs/node": "^8.3.4",
         "@auth/core": "^0.37.4",
-        "@genspectrum/dashboard-components": "^0.10.0",
+        "@genspectrum/dashboard-components": "0.10.1",
```
fengelniederhammer, line 24:
```suggestion
        "@genspectrum/dashboard-components": "^0.10.1",
```

website/src/components/rsv/RsvSingleVariantPage.astro
```diff
@@ -62,7 +63,15 @@ const lineageFilterConfigs = getLineageFilterConfigs(
...
-    <div class='flex flex-col gap-y-6' slot='dataDisplay'>
+    <div class='mx-[8px] flex flex-col gap-y-6' slot='dataDisplay'>
+        <div class='h-56 sm:h-32'>
+            <ComponentHeadline>Stats</ComponentHeadline>
+            <gs-statistics
+                numeratorFilter={JSON.stringify(variantFilter)}
+                denominatorFilter={JSON.stringify(baselineFilter)}
+                width='100%'
+                height='100%'></gs-statistics>
+        </div>
```
fengelniederhammer, line 74: I was thinking that maybe we could wrap those things in their own Astro (or React?) component (some styling div, the headline and the web component). In the end, we reuse the same blocks of code on every single variant page.

But I didn't think this through in full extent. We could do it separately somewhen later.

JonasKellerer: In general a good idea. I think we should do this after https://github.com/GenSpectrum/dashboards/pull/357 since I also refactored the views there a little.

website/package.json
JonasKellerer: Nice catch. It was a leftover from the local use.

PR #367, opened 2024-12-02
352 compare to baseline implement spike

### LINKED ISSUE #352: Compare to baseline: implement spike

JonasKellerer:
resolves #352

### Summary

This is currently only a preview. It shows a possible "compare to baseline" filter. It is similar to the "compare variants" page. I am unsure, if the baseline filter should be inside the "filter dataset" filter, or not.

I am looking forward to your feedback.

### Screenshot
grafik


JonasKellerer:
> Update: Ah so instead of comparing multiple variants to each other as in https://github.com/GenSpectrum/dashboards/pull/359/files would you now compare each added variant to a "baseline"? - In that case maybe I would choose to just add the baseline filters in the filter dataset - I think I would have found that clearer as a user

Yes. That is the use case. Thank you for your feedback. Maybe we also add to the page, that when no variant or baseline is selected, the user is shown a text “Please select …” like in #357. There we could also add some description.

https://cov-spectrum.org/explore/Germany/AllSamples/Past6M/variants?nextcladePangoLineage=XEC*&analysisMode=CompareToBaseline&

JonasKellerer:
As discussed: 
We keep the design like it is now. The baseline filter, is not only to filter the whole dataset, but also to compare the other variants to this. Thus, there can be prevalences over 100% in the prevalence over time plot. This allows to compare variants, which have no overlap.
To make this clearer, we decided to add a tooltip to the baseline filter. Also we add a description on the page, when there are no filters selected.

website/src/components/pageStateSelectors/VariantSelector.tsx
fengelniederhammer, line 1: `MutationFilterConfig` in this file seems unused.

website/src/components/pageStateSelectors/CompareVariantsToBaselineStateSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+        return {
+            ...pageState,
+            datasetFilter: {
+                location,
+                dateRange,
+            },
+            variants,
+            baselineFilter: toVariantFilter(baselineFilterConfigState),
+        };
```
fengelniederhammer, line 57: we set all keys anyway, we don't need the old page state.
```suggestion
    
}: {
    locationFilterConfig: LocationFilterConfig;
    dateRangeFilterConfig: DateRangeFilterConfig;
    baselineFilterConfig: VariantFilterConfig;
    variantFilterConfigs: Map<Id, VariantFilterConfig>;
    organismViewKey: OrganismViewKey & `${string}.${typeof compareToBaselineViewKey}`;
    organismsConfig: OrganismsConfig;
    
}) {
    const [location, setLocation] = useState<LapisLocation>(locationFilterConfig.initialLocation);
    const [dateRange, setDateRange] = useState<DateRangeOption>(dateRangeFilterConfig.initialDateRange);
    const [baselineFilterConfigState, setBaselineFilterConfigState] =
        useState<VariantFilterConfig>(baselineFilterConfig);

    const [variantConfigs, setVariantConfigs] = useState<Map<Id, VariantFilterConfig>>(variantFilterConfigs);

    const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);

    const newPageState = useMemo(() => {
        const variants = new Map(
            Array.from(variantConfigs).map(([id, variantFilterConfig]) => {
                return [id, toVariantFilter(variantFilterConfig)];
            }),
        );

        return {
            
            datasetFilter: {
                location,
                dateRange,
            },
            variants,
            baselineFilter: toVariantFilter(baselineFilterConfigState),
        };
```

website/src/components/pageStateSelectors/CompareVariantsToBaselineStateSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+    pageState: CompareVariantsData;
+}) {
+    const [location, setLocation] = useState<LapisLocation>(locationFilterConfig.initialLocation);
+    const [dateRange, setDateRange] = useState<DateRangeOption>(dateRangeFilterConfig.initialDateRange);
+    const [baselineFilterConfigState, setBaselineFilterConfigState] =
+        useState<VariantFilterConfig>(baselineFilterConfig);
+
+    const [variantConfigs, setVariantConfigs] = useState<Map<Id, VariantFilterConfig>>(variantFilterConfigs);
+
+    const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
```
fengelniederhammer, line 40:
```suggestion
    const view = useMemo(
        () => new Routing(organismsConfig).getOrganismView(organismViewKey),
        [organismsConfig, organismViewKey],
    );
```

website/src/components/pageStateSelectors/CompareVariantsToBaselineStateSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    onVariantFilterChange={setBaselineFilterConfigState}
+                    variantFilterConfig={baselineFilterConfigState}
+                />
+            </div>
+            <div>
+                <SelectorHeadline>Variant Filters</SelectorHeadline>
+                <VariantsSelector
+                    variantFilterConfigs={variantConfigs}
+                    setVariantFilterConfigs={setVariantConfigs}
+                    emptyVariantFilterConfig={view.pageStateHandler.getEmptyVariantFilterConfig()}
```
fengelniederhammer, line 83: I think we can simply things a bit by making this a callback:
```suggestion
                    emptyVariantFilterConfigProvider={view.pageStateHandler.getEmptyVariantFilterConfig}
```
Then we can get rid of `copyVariantFilterConfig`, because the provider always gets us a new object already anway.

fengelniederhammer (approved):
Nice :)

PR #369, opened 2024-12-02
docs(website): add contributing docs

JonasKellerer:
This is a starting point of a contributing readme. Feel free to add or remove points.

### DIFF
--- a/website/CONTRIBUTING.md
+++ b/website/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# Contributing to Dashboards
+
+Contributions are very welcome! Just fork the repository, develop in a branch and submit a pull request.
+
+## Commit Messages
+
+We follow [conventional commits](https://www.conventionalcommits.org) when writing commit messages.
+The messages themselves should help future developers understand **why** changes were made.
+
+## Code Style
+
+We value clean code, here are some guidelines on what we consider clean code:
+
+- Use expressive names (variables, classes, function), avoid abbreviations unless they are really common sense
+- Small functions and classes
+- Reasonable test coverage: We don't require 100% test coverage, but we do require tests for all non-trivial code.
+  Every piece of the code should at least be tested in the happy path.
+- Develop code with testability in mind.
+
+### Frontend
+
+We make use of some major libraries: Astro, React, Daisyui, and Tailwindcss.
+
+- use Astro when possible (for performance reasons)
+    - however avoid script tags, since they are harder to maintain. In this case use React.
+- when possible use css over js for interactivity
+    - it often yiels less and simpler code
+    - we can profit from browser parallelization and optimizations (modern browsers are highly optimized for parsing
+      HTML and CSS - when using
+      JS that's not immediately possible)
+- use React for interactive components
+- Use constant collection instead of strings for constants, eg.:
+    ```typescript
+    export const constantCollection = {
+        constant1: 'constant1' as const,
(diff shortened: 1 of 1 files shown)


website/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+-   Use expressive names (variables, classes, function), avoid abbreviations unless they are really common sense
+-   Small functions and classes
+-   Reasonable test coverage: We don't require 100% test coverage, but we do require tests for all non-trivial code.
+    Every piece of the code should at least be tested in the happy path.
+-   Develop code with testability in mind.
+
+### Frontend
+
+We make use of two frameworks: Astro, React, and Tailwindcss.
```
fengelniederhammer, line 22: It reads a bit funny, because it says "two", but then "Astro, React, and Tailwindcss" are 3. ;)
* Do we want to add DaisyUI to the list? Or do we consider it part of Tailwind for this purpose?
* I don't know whether framework is the right word here, but listing those is good, because it's the main ingedients to the website.

website/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+-   use Astro when possible (for performance reasons)
+    -   however avoid script tags, since they are harder to maintain
```
fengelniederhammer, line 25: Let's add something like "use React instead if you need client-side JS"?

website/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+-   when possible use css over js for interactivity
```
fengelniederhammer, line 26: I think we could add why:
* it often yields less and simpler code
* we can profit from browser optimizations (modern browsers are highly optimized for parsing HTML and CSS - when using JS that's not immediately possible)
* a lot of things (modals, overlays, etc) can be done with CSS only using standardized solutions (and add a reference to DaisyUI because it has many example how to do it?)

fengelniederhammer (commented):
Good start :+1:

website/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+-   use React for interactive components
+-   Use typed enums instead of strings for constants
+
+### Styling
+
+We use [Tailwind CSS](https://tailwindcss.com/) for styling.
+
+-   if possible use tailwind instead of custom css and inline styles
+-   consider using components for styling that is used in multiple places (or to improve readability)
```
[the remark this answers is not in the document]
JonasKellerer: I added an example and added the link.

website/CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Frontend
+
+We make use of two frameworks: Astro, React, and Tailwindcss.
+
+-   use Astro when possible (for performance reasons)
+    -   however avoid script tags, since they are harder to maintain
+-   when possible use css over js for interactivity
+-   use React for interactive components
+-   Use typed enums instead of strings for constants
```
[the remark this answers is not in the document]
JonasKellerer: I meant the constants like `dataOrigins`. I changed the wording to avoid the name enum.

website/CONTRIBUTING.md
JonasKellerer: I added daisyui to the list. It is one of our major libarries.

fengelniederhammer (approved):
LGTM

PR #370, opened 2024-12-02
chore(backend): use List instead of Array

fengelniederhammer:
### Summary

IntelliJ warned:
"Property with 'Array' type in a 'data' class: it is recommended to override 'equals()' and 'hashCode()'" This is the easiest way to get rid of that warning.


anna-parker (approved):
LGTM!

PR #371, opened 2024-12-02
refactor(website): make the "Apply filters" button a link

fengelniederhammer:
### Summary

So that the browser displays a preview where it's going when hovering it. And previously the button was essentially a link, because onChange just set `window.location.href`. We can let the `<a>` tag do that.

### Screenshot

grafik


JonasKellerer (approved):
LGTM

PR #382, opened 2024-12-04
feat(website): show message to select variant when none selected on compare variants page

### LINKED ISSUE #376: Show description when no variant filters are selected on compare variants page
On the compare variants page, show a message that the user has to select variants to compare. This is similar to #350. I would suggest, that the user has to select at least 2 variants.

fengelniederhammer:
resolves #376

### Summary

I also added some basic playwright tests for the page.

### Screenshot

grafik


JonasKellerer (approved):
LGTM.

The e2e test is a nice start.

I also wonder, if it would be good to provide some documentation about the view on the `SelectVariants` element. Not this PR, but maybe we can add a text, what this view should do. But maybe we find also another place on the page, where users can find more information on the view.

fengelniederhammer:
> maybe we can add a text, what this view should do. But maybe we find also another place on the page, where users can find more information on the view.

I'll add a discussion point to our next weekly.

PR #384, opened 2024-12-05
refactor(website): only pass organism - let the generic view page pick the view key

fengelniederhammer:


JonasKellerer (approved):
Good idea. LGTM.

PR #385, opened 2024-12-06
feat(website): add fallback element for filters

### LINKED ISSUE #363: Add fallback to client only elements
When using `client:only='react'` inside an astro component, astro skips the rendering step and the element is only put in, after the client has finished rendering it. However, this results in a page shift on the pages, where we have the StateSelectors (which are only loaded clientside). 

I found, that we could use https://docs.astro.build/en/reference/directives-reference/#clientonly to render a fallback. In this case we could create an element, which is as large as the coresponding element. This would then pervent the page shift. 

grafik

Nothing too urgent, but we can think about it.

JonasKellerer:
Resolves #363

### Summary

Adds a fallback to our client-side loaded PageStateSelectors. The page now show not shift anylonger. 

Know issues:
  - the mutation filter expands when there are mutations selected. It is hard to compensate for that
  - the covid collections are of different size. They therefore shift the other elements. However, on the analyze single variant page, this does not show so prominent, since the filter covers almost the whole screen on mobile.

### Screenshot

While loading:
grafik

When finished loading:
grafik

### DIFF
--- a/website/src/components/pageStateSelectors/FallbackElement.tsx
+++ b/website/src/components/pageStateSelectors/FallbackElement.tsx
@@ -1,0 +1,38 @@
+export function SequencingEffortsSelectorFallback() {
+    return <div className='h-[208px] lg:h-auto'></div>;
+}
+
+export function AnalyzeSingleVariantSelectorFallback() {
+    return <div className='h-[372px] lg:h-auto'></div>;
+}
+
+export function CompareSideBySideSelectorFallback({ numFilters }: { numFilters: number }) {
+    return (
+        <div>
+            {numFilters > 2 && <div className='h-[264px]' />}
+            {numFilters <= 2 && <div className='h-[220px]' />}
+        </div>
+    );
+}
+
+export function CompareToBaselineSelectorFallback({ numFilters }: { numFilters: number }) {
+    return (
+        <>
+            <div className='h-[460px] lg:h-auto'></div>
+            {Array.from({ length: numFilters }).map((_, index) => (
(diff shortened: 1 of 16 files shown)


website/src/components/pageStateSelectors/FallbackElement.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function AnalyzeSingleVariantSelectorFallback() {
+    return <div className='h-[372px] lg:h-auto'></div>;
+}
+
+export function CompareSideBySideSelectorFallback({ numFilters }: { numFilters: number }) {
+    return (
+        <div>
+            {numFilters > 2 && <div className='h-[264px]' />}
+            {numFilters <= 2 && <div className='h-[220px]' />}
```
fengelniederhammer, line 13: Why does this actually depend on `numFilters`?

fengelniederhammer (approved):
This is a nice improvement :+1:

website/src/components/pageStateSelectors/FallbackElement.tsx
JonasKellerer: The date range selector changes it height: For 2 filters it is in one row and for more filters it switches to the two row design.

JonasKellerer: Or is this only the case for me?

fengelniederhammer: I didn't check in detail, but the layout here is difficult anyway. Also the mutation filter will change its height depending on how many mutations are selected. Anything close to the actual value is good enough IMO.

PR #401, opened 2024-12-10
395 bring back mpox pages

JonasKellerer:
Resolves: #395

### Summary

Adds Mpox back to the list of organisms, with all views. Now uses the same sequencing efforts page as west-nile.

### DIFF
--- a/website/src/pages/mpox/index.astro
+++ b/website/src/pages/mpox/index.astro
@@ -1,0 +1,7 @@
+---
+import { ServerSide } from '../../views/serverSideRouting';
+
+return Astro.redirect(
+    ServerSide.routing.getOrganismView('mpox.singleVariantView').pageStateHandler.getDefaultPageUrl(),
+);
+---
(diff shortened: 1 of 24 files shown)


JonasKellerer:
@chaoran-chen What lineages do we want to display at the start of the "compare side-by-side" page? I now used no lineage and "F.2" which has the most counts in the last year.
I also used the same elements on the "sequencing efforts" page as for west-nile. Or do we want something different?

website/src/pages/mpox/sequencing-efforts.astro
fengelniederhammer, line 1: I think it's time for a `GenericSequencingEffortsPage` (not in this PR though) ;)

fengelniederhammer (approved):

website/src/pages/mpox/index.astro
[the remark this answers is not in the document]
JonasKellerer: Fixed it. It was also present on all other organism pages.

website/src/pages/mpox/sequencing-efforts.astro
JonasKellerer: True. I thought about it, but was not sure if there are some slight differences. I add a ticket.

Issue #402, opened 2024-12-10
Mega menu is not scrollable on small screens


JonasKellerer:
When on the two row design of the mega menu, and there are many entries, it is not possible to select elements at the bottom. 
Make it, so that the mega menu is scrollable, or that all entries are shown.

![grafik](https://github.com/user-attachments/assets/340efccf-585b-40f4-8a91-b31a16306de1)

PR #403, opened 2024-12-10
fix(website): allow scrolling on mega menu

JonasKellerer:
Resolves: #402

### Summary

Now allows for a scrollable mega menu. Also adjust the columns of the mega menu.

### Screenshot

Three column menu
grafik

Two column menu
grafik

### DIFF
--- a/website/src/layouts/base/header/MegaMenu.tsx
+++ b/website/src/layouts/base/header/MegaMenu.tsx
@@ -5,7 +5,9 @@ import type { WithClassName } from '../../../types/WithClassName.ts';
 export function MegaMenu({ className, children }: PropsWithChildren<WithClassName>) {
     return (
         <div className={`absolute left-0 w-screen bg-white shadow-lg ${className}`}>
-            <ul className='grid grid-cols-2 gap-4 border-2 border-gray-100 p-8 xl:grid-cols-5'>{children}</ul>
+            <ul className='grid max-h-[80vh] grid-cols-2 gap-4 overflow-y-scroll overscroll-none border-2 border-gray-100 p-8 lg:grid-cols-3 xl:grid-cols-5'>
+                {children}
+            </ul>
         </div>
     );
 }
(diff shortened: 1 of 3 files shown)


website/src/layouts/base/header/MegaMenu.tsx
```diff
@@ -5,7 +5,9 @@ import type { WithClassName } from '../../../types/WithClassName.ts';
 export function MegaMenu({ className, children }: PropsWithChildren<WithClassName>) {
     return (
         <div className={`absolute left-0 w-screen bg-white shadow-lg ${className}`}>
-            <ul className='grid grid-cols-2 gap-4 border-2 border-gray-100 p-8 xl:grid-cols-5'>{children}</ul>
+            <ul className='grid max-h-[80vh] grid-cols-2 gap-4 overflow-y-scroll border-2 border-gray-100 p-8 lg:grid-cols-3 xl:grid-cols-5'>
```
fengelniederhammer, line 8: I just found this: https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
Seems like a good idea to apply here
```suggestion
            <ul className='overscroll-none grid max-h-[80vh] grid-cols-2 gap-4 overflow-y-scroll border-2 border-gray-100 p-8 lg:grid-cols-3 xl:grid-cols-5'>
```

fengelniederhammer (approved):

website/src/views/ViewConstants.ts
```diff
@@ -14,9 +14,10 @@ export const singleVariantViewConstants = {
     iconType: 'magnify',
 } as const satisfies ViewConstants;
 
+const nonBreakingHyphen = '\u2011';
```
fengelniederhammer, line 17: I think your fancy hyphen broke the e2e tests ;) They are probably searching for a normal `-`.

website/src/layouts/base/header/MegaMenu.tsx
JonasKellerer: Good idea.

fengelniederhammer: (Actually it would be good to apply that to the `overlayThatClosesMenuOnClickOutside`, too, somehow, but I did not find a way to quickly do that. Let's keep it in mind for later. Doesn't matter much, but it would be nice if scrolling on the overlay did not scroll the page in the background)

PR #412, opened 2024-12-11
refactor(website): extract GenericSequencingEffortsPage

### LINKED ISSUE #406: Add GenericSequencingEfforts page
We now have two organism (west-nile, mpox) using the same page. Extract that page, similar to the the other "Generic" pages.

fengelniederhammer:
resolves #406

### Summary

As a preparation for #396...


website/src/views/OrganismConstants.ts
```diff
@@ -9,12 +9,24 @@ export interface OrganismConstants {
     readonly dataOrigins: DataOrigin[];
 }
 
+export const ComponentHeight = {
+    large: '600px',
+    small: '300px',
+} as const;
```
JonasKellerer, line 15: Maybe we should have a section on the page with smaller and larger elements, or refactor it, that the elements can have different sizes. (Not this PR)

website/src/components/views/sequencingEfforts/GenericSequencingEffortsPage.astro
```diff
@@ -91,24 +90,18 @@ const { label: subdivisionLabel, field: subdivisionField } = getLocationSubdivis
                 width='100%'
```
JonasKellerer, line 90: We could use the new constant here. 

Also I think, this can be the small size.

website/src/views/OrganismConstants.ts
JonasKellerer: We should also use these constants on the other "Generic" pages.

website/src/views/westNile.ts
```diff
@@ -46,14 +46,49 @@ class WestNileConstants implements SingleVariantConstants {
...
+            { label: 'Collection method', fields: ['collectionMethod'], height: ComponentHeight.small },
+            { label: 'Purpose of sampling', fields: ['purposeOfSampling'], height: ComponentHeight.small },
+            { label: 'Sample type', fields: ['sampleType'], height: ComponentHeight.small },
+            {
+                label: 'Amplicon PCR primer scheme',
+                fields: ['ampliconPcrPrimerScheme'],
+                height: ComponentHeight.small,
+            },
+            { label: 'Sequencing protocol', fields: ['sequencingProtocol'], height: ComponentHeight.small },
+        ];
```
JonasKellerer, line 89: I think we could extract this function to use it also for mpox.

website/tests/sequencingEfforts.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    [Organisms.westNile]: { location: 'USA' },
+    [Organisms.rsvA]: { location: 'USA' },
+    [Organisms.rsvB]: { location: 'USA' },
+    [Organisms.mpox]: { location: 'USA' },
+};
+
+test.describe('The Sequencing Efforts Page', () => {
+    for (const organism of allOrganisms) {
+        test.describe(organism, () => {
+            test(`should show diagrams after selecting two variants`, async ({ sequencingEffortsPage }) => {
```
JonasKellerer, line 18: This test name seems to be a copy relict.

website/tests/ViewPage.ts
JonasKellerer, line 1: I just thought, we could also check, that there are now errors texts on the pages.

JonasKellerer (changes requested):
Great to have a reusable view.

We can use the new height constant on more places :)

website/src/views/OrganismConstants.ts
fengelniederhammer: > We should also use these constants on the other "Generic" pages.

This will be part of my next PR, so I'll ignore this here.

website/src/components/views/sequencingEfforts/GenericSequencingEffortsPage.astro
fengelniederhammer: Reusing the constants will be part of my next PR anyway.

Small size is too small for some organisms (H5N1 has a lot of hosts).

JonasKellerer (approved):
LGTM.

PR #417, opened 2024-12-12
refactor(website): extract Gs*.astro components that wrap the web components

fengelniederhammer:
### Summary

Another preparation for #396 

This requires some types from the components library so that we can type the wrapper props.
Also clean up some types in general.


JonasKellerer (approved):
Nice. Makes it much easier to read.

PR #419, opened 2024-12-12
feat(website): Allow each organism to not include all pages

anna-parker:
resolves #

### Summary

Currently the generic view pages will error if an organism does not have that page - as the pages an organism has should be customizable we create a new organism subtype which includes all organisms with that specific view/page. 

Thanks @chaoran-chen for helping with this!

### DIFF
--- a/website/.prettierrc
+++ b/website/.prettierrc
@@ -1,18 +1,18 @@
 {
-  "printWidth": 120,
-  "tabWidth": 4,
-  "trailingComma": "all",
-  "semi": true,
-  "jsxSingleQuote": true,
-  "singleQuote": true,
-  "quoteProps": "consistent",
-  "plugins": ["prettier-plugin-astro", "prettier-plugin-tailwindcss"],
-  "overrides": [
-    {
-      "files": "*.astro",
-      "options": {
-        "parser": "astro"
-      }
-    }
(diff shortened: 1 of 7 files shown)


website/src/types/Organism.ts
```diff
@@ -67,6 +67,7 @@ export const organismConfig = {
 };
 export const allOrganisms = Object.keys(organismConfig) as Organism[];
 export type Organism = keyof typeof organismConfig;
+export type OrganismsWithViewKey<T, ViewKey extends string> = T extends `${infer O}.${ViewKey}` ? O : never;
```
fengelniederhammer, line 70: Let's put this in `routing.ts` (next to the definition of `OrganismViewKey`)

fengelniederhammer (approved):
Nice :)

website/src/types/Organism.ts
chaoran-chen: I think that it would be more consistent to write `Organism` as singular.

```suggestion
export type OrganismWithViewKey<T, ViewKey extends string> = T extends `${infer O}.${ViewKey}` ? O : never;
```

anna-parker: Done :-)

Issue #421, opened 2024-12-13
Comparison to baseline - Prevalence over time is broken on staging


anna-parker:
![image](https://github.com/user-attachments/assets/51ba4af3-a19a-42c3-88ef-aa723d8b23b9)

error reads: Unexpected value '[]' for "numeratorFilters": Array must contain at least 1 element(s)

fengelniederhammer:
It works when you add a variant - the fix should be that we add another text here that says "select a variant on the left" instead of the diagram if none is selected.

PR #422, opened 2024-12-13
feat(website): add compare to baseline landing page

anna-parker:
resolves https://github.com/GenSpectrum/dashboards/issues/421

### Summary

Since the refactor a landing page is required for the compare baseline page - this PR adds a landing page with a description of the component to be shown when no variants are selected.
image

It additionally adds an e2e test to ensure no component can have an error on a page on first load. Confirmed this fails on the main page currently: 
image

### DIFF
--- a/website/src/components/views/compareToBaseline/SelectBaseline.astro
+++ b/website/src/components/views/compareToBaseline/SelectBaseline.astro
@@ -1,0 +1,18 @@
+---
+import { PageHeadline } from '../../../styles/containers/PageHeadline';
+---
+
+<div class='mx-4 mt-20 flex min-h-64 flex-col items-center'>
+    <PageHeadline>Analyze a variant compared to a baseline</PageHeadline>
+    <div class='max-w-xl'>
+        <p class='mb-2'>To proceed, please select a variant filter.</p>
+        <p>
+            If no baseline is selected all sequences will be used as a baseline for comparing against sequences with a
+            specified variant.{' '}
+            Note that the dataset filter is applied to the entire dataset including both the baseline and variant subsets.{
+                ' '
+            }
+            Adjust the dataset filter as needed to refine your selection.
+        </p>
+    </div>
+</div>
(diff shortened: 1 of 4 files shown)


chaoran-chen (approved):
Looks great to me, thank you for finding and fixing the issue!

website/tests/mainPage.spec.ts
```diff
@@ -37,6 +41,8 @@ test.describe('Main page', () => {
                     .locator('..')
                     .getByText(linkName, { exact: true })
                     .click();
+                await expect(page.locator('text=Error -')).not.toBeVisible();
+                await expect(page.locator('text=Something went wrong')).not.toBeVisible();
```
fengelniederhammer, line 45: Doesn't `getByText` work here instead of `locator(...)`?

website/src/components/views/compareToBaseline/SelectBaseline.astro
```diff
@@ -0,0 +1,18 @@
...
+<div class='mx-4 mt-20 flex min-h-64 flex-col items-center'>
+    <PageHeadline>Analyze a variant compared to a baseline</PageHeadline>
+    <div class='max-w-xl'>
+        <p class='mb-2'>To proceed, please select a variant filter.</p>
+        <p>
+            If no baseline is selected all sequences will be used as a baseline for comparing against sequences with a
+            specified variant.{' '}
+            Note that the dataset filter is applied to the entire dataset including both the baseline and variant subsets.{
+                ' '
+            }
```
fengelniederhammer, line 14: Those whitespaces look funny. Do we need them at all?

website/tests/mainPage.spec.ts
anna-parker: My linter (prettier) complained and suggested this - it should do the same thing and getByTest makes some type assumptions that are apparently not accounted for in the tests

website/src/components/views/compareToBaseline/SelectBaseline.astro
anna-parker: Ah true! Good catch - they aren't needed in astro - prettier formatted them weirdly - I have now removed :-)

PR #423, opened 2024-12-13
feat(website): add option to search by clades on sequencing efforts page

anna-parker:
This will be very useful for influenza - but its anyways nice to give users more filter options - this also gives the option to not show sequence mutations on the compare variant page - which will also be useful for Influenza and other parent pages where we group all information and do not align. 

resolves #

### Screenshot

https://github.com/user-attachments/assets/04ed1a51-3c95-426a-82b2-004ff4f7fbc9

### DIFF
--- a/website/src/components/pageStateSelectors/SequencingEffortsPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/SequencingEffortsPageStateSelector.tsx
@@ -4,6 +4,8 @@ import { useMemo, useState } from 'react';
 import { ApplyFilterButton } from './ApplyFilterButton.tsx';
 import { BaselineSelector, type DateRangeFilterConfig, type LocationFilterConfig } from './BaselineSelector.tsx';
 import { SelectorHeadline } from './SelectorHeadline.tsx';
+import { toVariantFilter, type VariantFilterConfig } from './VariantFilterConfig.ts';
+import { VariantSelector } from './VariantSelector.tsx';
 import type { OrganismsConfig } from '../../config.ts';
 import type { LapisLocation } from '../../views/helpers.ts';
 import { type OrganismViewKey, Routing } from '../../views/routing.ts';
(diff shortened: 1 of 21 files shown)


website/src/components/pageStateSelectors/SequencingEffortsPageStateSelector.tsx
```diff
@@ -45,6 +51,14 @@ export function SequencingEffortsPageStateSelector({
                     dateRangeFilterConfig={dateRangeFilterConfig}
                 />
             </div>
+            <div>
+                <SelectorHeadline>Variant Filter</SelectorHeadline>
```
fengelniederhammer, line 55: Do we need this headline? I think on this page it makes more sense to only have "filters", because we don't differentiate the filters in the diagrams anymore.

On the other pages this separation is useful, because the dataset filters are a baseline and the variant filters acts on top of it (e.g. for computing the prevalence over time).
```suggestion
```

website/src/components/pageStateSelectors/VariantSelector.tsx
```diff
@@ -6,9 +6,11 @@ import { GsMutationFilter } from '../genspectrum/GsMutationFilter.tsx';
 export function VariantSelector({
     onVariantFilterChange,
     variantFilterConfig,
+    unaligned,
```
fengelniederhammer, line 9: I don't like the name of this prop - I would rather call it something like "hide mutations filter" or similar. Or the version that doesn't require double negations (with a `true` default value):

```suggestion
    showMutationFilter = true,
```

website/src/components/views/compareSideBySide/GenericCompareSideBySidePage.astro
```diff
@@ -15,9 +15,10 @@ import { CompareSideBySideSelectorFallback } from '../../pageStateSelectors/Fall
 type OrganismViewCompareVariant = OrganismWithViewKey<typeof compareSideBySideViewKey>;
 interface Props {
     organism: OrganismViewCompareVariant;
+    unaligned?: boolean;
```
fengelniederhammer, line 18: Is this prop actually set anywhere?

website/src/views/covid.ts
```diff
@@ -294,6 +296,10 @@ export class CovidSequencingEffortsView extends BaseView<
                         location: {},
                         dateRange: constants.defaultDateRange,
                     },
+                    variantFilter: {
+                        lineages: {},
+                        mutations: {},
```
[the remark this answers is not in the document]
fengelniederhammer: But it's probably ok for now - at least the implementation is less effort than if we would introduce another filter type.

fengelniederhammer (approved):

website/src/views/covid.ts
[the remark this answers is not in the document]
anna-parker: I'm not sure I fully understand you - do you mean that the filter is labeled "Variant filter", or is this visible somewhere else that I am missing? Otherwise, I agree - it would be great to be able to customize the label of the Variant Filter for each page.

website/src/components/views/compareSideBySide/GenericCompareSideBySidePage.astro
anna-parker: I use it later for Influenza - but not yet

fengelniederhammer: Should we remove it for now as it's currently still unused?

website/src/views/covid.ts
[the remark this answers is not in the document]
fengelniederhammer: I mean: The type has `mutations`, but they will never be used on sequencing efforts pages, because we don't have a mutations filter there.

PR #427, opened 2024-12-16
396 add gs sequences by location to sequencing efforts

### LINKED ISSUE #396: Add gs-sequences-by-location to sequencing efforts

fengelniederhammer:
resolves #396

### Summary

* Adding topo json maps that the Astro server can serve at `mapData/{locationName}.topo.json`
  * Currently available: World, Germany, Switzerland, USA
* Replacing the `gs-aggregate` table for the geographic distribution by `gs-sequences-by-location` on the sequencing efforts pages. The map will be shown if the user filters for a location which has a map.

Should we hide maps on pages that don't have good enough data? And manually enable them once the data is cleaned sufficiently?

### Screenshot

It works quite well for Covid:
grafik

grafik

The world map mostly works for the other organisms too, but the data isn't properly cleaned for country divisions:
grafik


website/src/util/getSequencesByLocationMapData.ts
```diff
@@ -0,0 +1,93 @@
...
+        offsetX: 10,
+        offsetY: 51.4,
+    },
+    Switzerland: {
+        zoom: 7.7,
+        offsetX: 8.2,
+        offsetY: 46.8,
+    },
+    USA: {
+        zoom: 4.1,
```
JonasKellerer, line 19: Just to be sure: we don't want to have a zoom level, where Alaska and Hawaii are shown? Or is there also another map available?

website/src/util/getSequencesByLocationMapData.ts
```diff
@@ -0,0 +1,93 @@
+import type { MapSource } from '@genspectrum/dashboard-components/util';
+import type { Topology } from 'topojson-specification';
+
+let knownMapFiles: ReturnType<typeof loadKnownMapFiles> | null = null;
+
+/* eslint-disable @typescript-eslint/naming-convention -- the keys must match the country names */
+const mapMetadata: Record<string, Omit<MapData, 'mapSource'>> = {
+    Germany: {
```
JonasKellerer, line 8: Is there a way to set the color of the boundaries to black? I just looked at covid sampling in Germany and it is hard to see the borders.
grafik

fengelniederhammer: Hm, true. Or should we use a lighter grey for locations without data?

website/src/util/getSequencesByLocationMapData.ts
fengelniederhammer: We agreed that for now it's good enough, especially because users can scroll themselves. We can tweak the map later if we think that it's necessary.

website/src/util/getSequencesByLocationMapData.ts
JonasKellerer: I would be fine with both. I have not tested it. It should just work with all our colors.

website/src/util/getSequencesByLocationMapData.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+};
+/* eslint-enable @typescript-eslint/naming-convention */
+
+function loadKnownMapFiles() {
+    const matches = import.meta.glob<Topology>('../../public/mapData/*.topo.json', { eager: true }); // For reference: https://vite.dev/guide/features.html#glob-import
+
+    const mapSpecs = Object.entries(matches).map(([relativeMapFilePath, map]) => {
+        const lastSlashIndex = relativeMapFilePath.lastIndexOf('/');
+        const fileEndingIndex = relativeMapFilePath.lastIndexOf('.topo.json');
```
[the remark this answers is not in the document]
fengelniederhammer: Nice, yes, works!

website/src/util/getSequencesByLocationMapData.ts
fengelniederhammer: I'll change it in the components.

Issue #432, opened 2024-12-17
Add download button to pages with a single variant


fengelniederhammer:
Add the download button to pages that only have a single variant first - then we can investigate later how we can do it for pages with multiple variants later.

PR #433, opened 2024-12-18
feat(website): add button to download accession list of the sequences used on each page

### LINKED ISSUE #425: Add button to download accessions of sequences
There should be a button for each view to download the accessions of all sequences that are used for the analysis (i.e., after applying the filters). This would just be a link to `/sample/details?fields=accession,insdcAccession&dataFormat=tsv`.

For non-pathoplexus data, people can only downlod the list of INSDC accession (not the Loculus genspectrum accessions)

This is particularly relevant for data from Pathoplexus which requires providing a list of accessions explicitly in the [terms](https://pathoplexus.org/about/terms-of-use/data-use-terms#421-unpublished-and-un-preprinted-work) for restricted-use data.

- [ ] #432

fengelniederhammer:
resolves #425, #432

### Summary

This adds a "download data" button to the footer of every organism page. Clicking it opens a modal that offers links to download the accessions of every variant on the page.

### Screenshot

grafik

### DIFF
--- a/website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
+++ b/website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
@@ -6,6 +6,7 @@ import { chooseGranularityBasedOnDateRange } from '../../../util/chooseGranulari
 import { hasOnlyUndefinedValues } from '../../../util/hasOnlyUndefinedValues';
 import { getLineageFilterConfigs, getLineageFilterFields } from '../../../views/View';
 import { getLocationSubdivision } from '../../../views/locationHelpers';
+import { toDisplayName } from '../../../views/pageStateHandlers/PageStateHandler';
 import { type OrganismViewKey, type OrganismWithViewKey } from '../../../views/routing';
 import { ServerSide } from '../../../views/serverSideRouting';
 import { singleVariantViewKey } from '../../../views/viewKeys';
@@ -17,6 +18,7 @@ import GsMutationsOverTime from '../../genspectrum/GsMutationsOverTime.astro';
 import GsPrevalenceOverTime from '../../genspectrum/GsPrevalenceOverTime.astro';
 import { AnalyzeSingleVariantSelectorFallback } from '../../pageStateSelectors/FallbackElement';
 import { SingleVariantPageStateSelector } from '../../pageStateSelectors/SingleVariantPageStateSelector';
+import { sanitizeForFilename } from '../compareSideBySide/toDownloadLink';
 
 type OrganismViewCompareVariant = OrganismWithViewKey<typeof singleVariantViewKey>;
 interface Props {
@@ -47,9 +49,20 @@ const lineageFilterConfigs = getLineageFilterConfigs(
 );
 
 const noVariantSelected = hasOnlyUndefinedValues(pageState.variantFilter);
+
+const displayName = toDisplayName(pageState.variantFilter);
+const downloadLinks = noVariantSelected
+    ? []
+    : [
+          {
+              label: `Download accessions of "${displayName}"`,
+              filter: variantLapisFilter,
+              downloadFileBasename: `${organism}_${sanitizeForFilename(displayName)}_accessions`,
+          },
+      ];
 ---
 
-<SingleVariantOrganismPageLayout view={view}>
+<SingleVariantOrganismPageLayout view={view} downloadLinks={downloadLinks}>
     <SingleVariantPageStateSelector
         slot='filters'
         locationFilterConfig={{
--- a/website/src/components/views/compareSideBySide/toDownloadLink.ts
+++ b/website/src/components/views/compareSideBySide/toDownloadLink.ts
@@ -1,0 +1,22 @@
+import type { Organism } from '../../../types/Organism.ts';
+import type { DatasetAndVariantData, Id } from '../../../views/View.ts';
+import type { CompareSideBySideStateHandler } from '../../../views/pageStateHandlers/CompareSideBySidePageStateHandler.ts';
+import { toDisplayName } from '../../../views/pageStateHandlers/PageStateHandler.ts';
+
+export function toDownloadLink(pageStateHandler: CompareSideBySideStateHandler, organism: Organism) {
+    return ([id, { variantFilter, datasetFilter }]: [Id, DatasetAndVariantData]) => {
+        const filter = pageStateHandler.variantFilterToLapisFilter(datasetFilter, variantFilter);
+        const displayName = toDisplayName(variantFilter);
+        const variantName = displayName.length > 0 ? displayName : `variant ${id + 1}`;
+
+        return {
+            label: `Download accessions of "${variantName}"`,
+            filter,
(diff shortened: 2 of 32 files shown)


website/src/components/views/sequencingEfforts/GenericSequencingEffortsPage.astro
```diff
@@ -42,7 +42,16 @@ const lineageFilterConfigs = getLineageFilterConfigs(
 );
 ---
 
-<SingleVariantOrganismPageLayout view={view}>
+<SingleVariantOrganismPageLayout
+    view={view}
+    downloadLinks={[
+        {
+            label: 'Download accessions',
+            filter: datasetLapisFilter,
```
JonasKellerer, line 50:
```suggestion
            filter: variantLapisFilter,
```
(to fix the error)
But I would prefer renaming the variable to "lapisFilter", since we don't show variants on the sequencing efforst pages.

website/src/components/views/compareSideBySide/toDownloadLink.ts
```diff
@@ -0,0 +1,21 @@
...
+
+        return {
+            label: `Download accessions of ${variantName}`,
+            filter,
+            downloadFileBasename: `${makeFilenameFriendly(variantName)}_accessions`,
+        };
+    };
+}
+
+export function makeFilenameFriendly(variantName: string) {
```
JonasKellerer, line 19: I hope the filename is nice to me ;)
But I'm not sure what this function does. :see_no_evil: Makes it the filename more readable, or better for all filesystems?

website/src/layouts/OrganismPage/AccessionsDownloadButton.astro
```diff
@@ -0,0 +1,61 @@
...
+                        label,
+                        url: assembleDownloadUrl(
+                            view.organismConstants.accessionDownloadFields,
+                            filter,
+                            downloadFileBasename,
+                            getLapisUrl(view.organismConstants.organism),
+                        ),
+                    }))
+                    .map(({ label, url }) => (
+                        <a class='link m-2 flex items-center' href={url}>
```
JonasKellerer, line 50:
```suggestion
                        <a class='link m-2 flex gap-1 items-center' href={url}>
```
To make the icon and the text clearly distinguishable.

website/src/layouts/OrganismPage/AccessionsDownloadButton.astro
```diff
@@ -0,0 +1,61 @@
...
+                            <span class='iconify mdi--download' />
```
JonasKellerer, line 51:
```suggestion
                        <a class='link m-2 flex items-center gap-1' href={url}>
                            <span class='iconify mdi--download text-xl' />
```
I would prefer, that the button has the same size as the capital "D" of "Download", and also align the underline with the underline of the link.

website/src/layouts/OrganismPage/AccessionsDownloadButton.astro
```diff
@@ -0,0 +1,61 @@
...
+    view: View<object, OrganismConstants, PageStateHandler<object>>;
+    downloadLinks: DownloadLink[];
+}
+
+const { view, downloadLinks } = Astro.props;
+
+const modalId = 'accessionDownloadModal';
+---
+
+<button class='mr-4 flex items-center p-2 text-xs' onclick={`${modalId}.showModal()`}>
```
JonasKellerer, line 29:
```suggestion
<button class='mr-4 flex items-center p-2 text-xs gap-1' onclick={`${modalId}.showModal()`}>
```

website/src/layouts/OrganismPage/OrganismPageLayout.astro
```diff
@@ -23,6 +27,7 @@ const lastUpdatedDate = await getLastUpdatedDate(getLapisUrl(view.organismConsta
     </div>
     <slot />
     <div class='flex justify-center' slot='secondary-footer'>
+        {downloadLinks.length > 0 && <AccessionsDownloadButton view={view} downloadLinks={downloadLinks} />}
         <span class='mr-4 p-2 text-xs'>Data last updated: {lastUpdatedDate ?? 'unknown'}</span>
```
JonasKellerer, line 31: Not really this PR, but it would be nice to put this line into its own component, to have the same complexity as the other lines.

website/src/layouts/OrganismPage/OrganismPageLayout.astro
```diff
@@ -23,6 +27,7 @@ const lastUpdatedDate = await getLastUpdatedDate(getLapisUrl(view.organismConsta
     </div>
     <slot />
     <div class='flex justify-center' slot='secondary-footer'>
```
JonasKellerer, line 29: I think the alignment is lightly off on mobile:
grafik
Maybe a combination of max width and justify-between might help. We could also get rid of the "mr-4" through a "gap-x".

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
```diff
@@ -47,9 +49,20 @@ const lineageFilterConfigs = getLineageFilterConfigs(
 );
 
 const noVariantSelected = hasOnlyUndefinedValues(pageState.variantFilter);
+
+const displayName = toDisplayName(pageState.variantFilter);
+const downloadLinks = noVariantSelected
+    ? []
+    : [
+          {
+              label: `Download accessions of ${displayName}`,
```
JonasKellerer, line 58:
```suggestion
              label: `Download accessions of "${displayName}"`,
```
To make it clear that "A123T + 2.34" is one variant

JonasKellerer: (Same for the other views)

JonasKellerer (changes requested):
Nice. I also like the modal rework. 
Mostly, I would favor some slight ui changes.

website/src/components/views/compareSideBySide/toDownloadLink.ts
fengelniederhammer: It makes the variant name filename-friendly. :D I'm not sure what to rename it to, do you have a suggestion?

website/src/layouts/OrganismPage/AccessionsDownloadButton.astro
fengelniederhammer: It's still not properly aligned, but better.

website/src/components/views/compareSideBySide/toDownloadLink.ts
JonasKellerer: Ah. It is one word :) 
What do you think about: "sanitizeForFilename", or "makeValidFilename"?

website/src/layouts/OrganismPage/OrganismPageLayout.astro
JonasKellerer: Stacking is a great idea. I think it looks good now and the point before the sm breakpoint is not that bad.

JonasKellerer (approved):

Issue #436, opened 2024-12-19
Some variants in collections don't work


fengelniederhammer:
When I try to select the "Other Recombinants" in the "#1 editors choice" collection on the Covid single variant page, nothing happens and the console shows:
```
Uncaught ZodError: [
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "nucMutations"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "aaMutations"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "nucInsertions"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
    "path": [
      "aaInsertions"
    ],
    "message": "Required"
  }
]
``` 

It also happens for some other collections.

PR #438, opened 2024-12-19
fix(website): show error when parsing the collection variant filter fails

fengelniederhammer:
### Summary

Preparation for https://github.com/GenSpectrum/dashboards/issues/436 - at least we can see the error now.

### Screenshot

grafik

grafik


fengelniederhammer:
> Only, when I try to copy the error using the "Copy error information to clipboard" button, nothing happens. I'm not sure, if this is introduced in this PR.

-> https://github.com/GenSpectrum/dashboards/pull/442

PR #442, opened 2024-12-20
fix(website): copy error to clipboard button

fengelniederhammer:


JonasKellerer (approved):
Ah. Sry. I think I introduced this bug myself. :see_no_evil:

PR #444, opened 2024-12-20
feat(website): add fragment links to scroll to component headlines

### LINKED ISSUE #426: Direct links to the visualization components / subtitles
As the views can be quite big (i.e., have a lot of components), it would be useful if we can have direct anchor links to the (visualization) components. Maybe with a UX similar to GitHub where you see a clickable link icon when you hover over a subtitle?

image

fengelniederhammer:
resolves #426

### Screenshot

Clicking the headline will scroll to the component. When hovering the headline, a link symbol is displayed.
grafik

### DIFF
--- a/website/src/components/ComponentHeadline.astro
+++ b/website/src/components/ComponentHeadline.astro
@@ -1,7 +1,17 @@
 ---
+import { fragmentLinkScrollMargin } from '../layouts/base/header/headerConstants';
+interface Props {
+    title: string;
+}
 
+const { title } = Astro.props;
+
+const fragmentId = title.toLowerCase().replace(/ /g, '-');
 ---
 
-<h2 class='font-bold'>
-    <slot />
+<h2 class={`font-bold ${fragmentLinkScrollMargin} `} id={fragmentId}>
+    <a href={`#${fragmentId}`} class='group flex items-center gap-1'>
+        {title}
+        <span class='iconify font-normal opacity-0 mdi--link-variant group-hover:opacity-100'></span>
+    </a>
 </h2>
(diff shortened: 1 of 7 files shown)


website/src/components/ComponentHeadline.astro
```diff
@@ -1,7 +1,17 @@
...
+}
 
+const { title } = Astro.props;
+
+const fragmentId = title.toLowerCase().replace(/ /g, '-');
 ---
 
-<h2 class='font-bold'>
-    <slot />
+<h2 class={`font-bold ${fragmentLinkHeaderOffset} `} id={fragmentId}>
```
JonasKellerer, line 12:
```suggestion
<h2 class={`font-bold scroll-mt-20`} id={fragmentId}>
```
I think this will do the same. We could also put this into the headerConstants, to see that it has the same height as the header. But maybe we could also leave it here.

JonasKellerer: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top

website/src/layouts/base/header/headerConstants.ts
```diff
@@ -1 +1,8 @@
 export const headerHeight = 'h-16';
+
+/**
+ * Account for the header when scrolling to a fragment link (href="#my-fragment").
+ * Otherwise, the fragment that we scroll to will be hidden behind the header.
+ * The height must match the header height (plus a little extra spacing).
+ */
```
JonasKellerer, line 7: See comment above.

JonasKellerer (changes requested):
Nice addition. I would only use the buildin css feature for the scroll margin.

website/src/components/ComponentHeadline.astro
fengelniederhammer: Nice, I didn't know of that feature!

JonasKellerer (approved):

PR #452, opened 2025-01-03
feat(website): update GsTextInput and GsLineageFilter to only show options that correspond to current filters

anna-parker:
resolves https://github.com/GenSpectrum/dashboard-components/issues/609, https://github.com/GenSpectrum/dashboard-components/issues/632

### Summary

Currently the search bar shows all clades and countries that have ever been sequenced regardless of which other filters (such as time) have been applied - this means that for our start page most options will lead to 0 results. This isn't great for user experience. This filters down the options list in each filter using the currently applied filters.

As we currently apply filters based on page state this now means each filter change will result in a page reload. I do not see another solution to this problem without wrapping the filter bar in a react component with another temporary state.

### Screenshot

https://github.com/user-attachments/assets/493f3339-9859-4417-9e17-80ff5a7d95ee

Relies on https://github.com/GenSpectrum/dashboard-components/pull/633

## TO DO
- [x] use for location filter as well
- [x] make sure that new lapisFilter is applied as soon as new options are selected and not only when apply filters button is clicked

### DIFF
--- a/website/src/components/genspectrum/GsLineageFilter.tsx
+++ b/website/src/components/genspectrum/GsLineageFilter.tsx
@@ -1,3 +1,4 @@
+import type { LapisFilter } from '@genspectrum/dashboard-components/util';
 import { useEffect, useRef } from 'react';
 
 import '@genspectrum/dashboard-components/components';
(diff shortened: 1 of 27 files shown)


JonasKellerer:
I'm a bit confused. The PR mentiones it resolves tickets from dashborad-components, but this PR is in dashboards. I think they should be independent from each other. 
Also I am not sure about the direction of this PR. I am not sure if I like, that the page reloads each time when a filter is changed. Maybe we find a better solution like you suggested, that only the filters refetch their autocomplete list, but we still have the "Filter" button.

anna-parker:
closing in favor of https://github.com/GenSpectrum/dashboards/pull/486

PR #480, opened 2025-01-20
feat: remove dev config

fengelniederhammer:
### Summary

It mirrors the staging environment anyway. I don't see a real use case anymore.


anna-parker (commented):
generally looks good to me - just a quick query.

1. so dev was only used for local development?
2. if dev==staging why switch from dev to prod and not dev to staging?

fengelniederhammer:
>     1. so dev was only used for local development?

I used it sometimes for local development. "dev" is not used on any proper deployment.

>     2. if dev==staging why switch from dev to prod and not dev to staging?

Prod is more stable than staging. Also we're not modifying any data, since LAPIS is read-only. Thus it's actually better to use prod while developing and in tests (IMO).

anna-parker (approved):

PR #497, opened 2025-01-23
feat(website): WISE dashboard - add influenza specific info

anna-parker:
resolves #

### Summary

This requires https://github.com/GenSpectrum/dashboard-components/pull/686

As requested prior to the launch tomorrow we need to add info to the dashboards - this adds info for influenza - sadly the links do not show as links as this has to be passed to the dashboard components

### DIFF
--- a/website/src/components/genspectrum/GsWastewaterMutationsOverTime.astro
+++ b/website/src/components/genspectrum/GsWastewaterMutationsOverTime.astro
@@ -7,9 +7,10 @@ interface Props {
     title: string;
     lapisFilter: LapisFilter;
     sequenceType: SequenceType;
+    info?: string;
 }
 
-const { title, lapisFilter, sequenceType } = Astro.props;
+const { title, lapisFilter, sequenceType, info } = Astro.props;
 ---
 
 <ComponentWrapper title={title}>
@@ -17,5 +18,8 @@ const { title, lapisFilter, sequenceType } = Astro.props;
         lapisFilter={JSON.stringify(lapisFilter)}
         sequenceType={sequenceType}
         width='100%'
-        height='100%'></gs-wastewater-mutations-over-time>
+        height='100%'
+    >
+        <span slot='infoText'><Fragment set:html={info} /></span>
+    </gs-wastewater-mutations-over-time>
 </ComponentWrapper>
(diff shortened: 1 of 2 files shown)


chaoran-chen:
> sadly the links do not show as links as this has to be passed to the dashboard components

We could consider accepting markdown which then gets rendered in the components. But maybe not relevant for now and can be added in the future?

website/src/pages/swiss-wastewater/flu.astro
```diff
@@ -42,6 +80,13 @@ const references = (await (await fetch(getReferencesUrl)).json()).data.map(
                                     isRevocation: false,
                                 }}
                                 sequenceType={sequenceType}
+                                info={
+                                    sequenceType === 'amino acid'
+                                        ? (additionalInfo.get(reference) ?? '') +
+                                          ' ' +
+                                          (additionalAAInfo.get(reference) ?? '')
+                                        : (additionalInfo.get(reference) ?? '')
```
chaoran-chen, line 88: Idea to simplify it a bit:

```suggestion
                                    (additionalInfo.get(reference) ?? '') + 
                                    (sequenceType === 'amino acid' ? ' ' + (additionalAAInfo.get(reference) ?? '') : '')
                                    
```

chaoran-chen (approved):
Thanks, @anna-parker!

website/src/components/genspectrum/GsWastewaterMutationsOverTime.astro
```diff
@@ -7,15 +7,19 @@ interface Props {
...
 
 <ComponentWrapper title={title}>
     <gs-wastewater-mutations-over-time
         lapisFilter={JSON.stringify(lapisFilter)}
         sequenceType={sequenceType}
         width='100%'
-        height='100%'></gs-wastewater-mutations-over-time>
+        height='100%'
+    >
+        <span slot='infoText'><Fragment set:html={info} /></span>
```
fengelniederhammer, line 23: Why do we need that `Fragment` here?

chaoran-chen: Currently, the info text is provided as a string with HTML (see [here](https://github.com/GenSpectrum/dashboards/pull/497/files#diff-42f81b60a3c67d368026866e4b38927feeaf1f3b72f29bcb60a80729048f24f7R38)), this way, we need a Fragment, don't we? Is there a better way?

fengelniederhammer: Ah I see... Why don't we use an Astro slot here?
```suggestion
        <span slot='infoText'><slot name="infoText" /></span>
```

And then, where the component is called, we can use it as 

```
<gs-wastewater...>
  <span slot="infoText">text... <a>important link</a></span>
</gs-wastewater...>
``

anna-parker: this didn't work btw - and I decided as we are under time constraints to merge without

PR #500, opened 2025-01-27
use an Astro slot instead of `set:html`

fengelniederhammer:

### DIFF
--- a/website/src/components/InfluenzaWastewaterInfo.astro
+++ b/website/src/components/InfluenzaWastewaterInfo.astro
@@ -1,0 +1,56 @@
+---
+import type { SequenceType } from '@genspectrum/dashboard-components/util';
+
+interface Props {
+    reference: string;
+    sequenceType: SequenceType;
+}
+
+const { reference, sequenceType } = Astro.props;
+
+/* eslint-disable @typescript-eslint/naming-convention */
+const additionalInfo = {
+    H1: {
+        referenceSequence: 'CY121680.1 (A/California/07/2009(H1N1))',
+        translationGene: 'HA1 + HA2 (HA0)',
+        genemapLink: 'https://github.com/nextstrain/seasonal-flu/blob/master/config/h1n1pdm/ha/genemap.gff',
+    },
+    N1: {
+        referenceSequence: 'CY121682.1 (A/California/07/2009(H1N1))',
+        translationGene: 'NA',
+        genemapLink: 'https://github.com/nextstrain/seasonal-flu/blob/master/config/h1n1pdm/ha/genemap.gff',
+    },
+    H3: {
+        referenceSequence: 'CY163680.1 (A/Wisconsin/67/2005(H3N2))',
+        translationGene: 'HA1 + HA2 (HA0)',
+        genemapLink: 'https://github.com/nextstrain/seasonal-flu/blob/master/config/h3n2/ha/genemap.gff',
+    },
+    N2: {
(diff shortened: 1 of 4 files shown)


anna-parker (approved):
Personally I think that hard-coding the structure of the information box isn't a great idea as the structure might still change in the future. But if you want to merge I won't block.

fengelniederhammer:
> Personally I think that hard-coding the structure of the information box isn't a great idea as the structure might still change in the future. But if you want to merge I won't block.

I'm aware of that. Maybe we need to change it back later. But for now it was easier to change it to that. Even if we need to change the structure later, that's an easy change. Let's see.

website/src/components/InfluenzaWastewaterInfo.astro
```diff
@@ -0,0 +1,56 @@
...
+</span>
+{
+    sequenceType === 'amino acid' && (
+        <span>
+            Same data as under Nucleotide mutations, but nucleotide substitutions are translated into amino acid
+            substitutions relative to {info.translationGene} gene, for genemap see{' '}
+            {/* prettier-ignore *//* make sure there is no space between the link and the dot */}
+            <a class="text-blue-600 hover:text-blue-800" href={info.genemapLink}>here</a>.
+        </span>
+    )
```
JonasKellerer, line 55: For the other components, we write the info in sentence style. Do we want to do the same here?

website/src/components/InfluenzaWastewaterInfo.astro
```diff
@@ -0,0 +1,56 @@
...
+<span>
+    Frequencies of all nonsynonymous mutations (minimum read depth 100x) in the IAV
+    {reference} segment relative to reference {info.referenceSequence}.
+</span>
+{
+    sequenceType === 'amino acid' && (
+        <span>
+            Same data as under Nucleotide mutations, but nucleotide substitutions are translated into amino acid
+            substitutions relative to {info.translationGene} gene, for genemap see{' '}
+            {/* prettier-ignore *//* make sure there is no space between the link and the dot */}
```
JonasKellerer, line 52: I think adding a new span will do the trick:

```
{
    sequenceType === 'amino acid' && (
    <span>
            Same data as under Nucleotide mutations, but nucleotide substitutions are translated into amino acid
            substitutions relative to {info.translationGene} gene, for genemap see
            </span>
    <span>
    <a class="text-blue-600 hover:text-blue-800" href={info.genemapLink}>here</a>.
                </span>
        )
    }
```

website/src/components/InfluenzaWastewaterInfo.astro
fengelniederhammer: What do you mean by "sentence style" here? Can you give an example?

JonasKellerer: For the mutations over time component, we write the description in sentences: "This component displays mutations (substitutions and deletions) over time for a dataset selected by a LAPIS filter." ...

So in this case we could write: 
"This component shows the frequenciesof all nonsynonymous mutations (minimum read depth 100x) in the IAV
    {reference} segment relative to reference {info.referenceSequence}."
    
"It uses the same data as, ... "

fengelniederhammer: Ah you mean full sentences instead of bullet point form? I just took the sentences as they were.

@anna-parker What do you think? Should we change that? I don't really have a preference here.

PR #509, opened 2025-01-29
Draft for more filters

JonasKellerer:
### Summary

In this draft, the new design of the filters is shown. The design is used for all pages, but only on the `/west-nile/single-variant` page, it is connected to config values.
Before updating also the other views, it would be nice to get some feedback on the implementation and design. 

Possible future extensions:
  - add color to baseline and variant filter in conjuction wtih a describing grafic wtih the same colors 
 
After mockup:
  - add functionality to "Reset" button
  - add functionality to "Add filter fields" button
  - add functionality to "Help" button
  - specify the filter fields in config 

### Screenshot

grafik

### DIFF
--- a/website/src/components/pageStateSelectors/BaselineSelector.tsx
+++ b/website/src/components/pageStateSelectors/BaselineSelector.tsx
@@ -3,46 +3,144 @@ import type { DateRangeOption } from '@genspectrum/dashboard-components/util';
 import type { LapisLocation } from '../../views/helpers.ts';
 import { GsDateRangeSelector } from '../genspectrum/GsDateRangeSelector.tsx';
 import { GsLocationFilter } from '../genspectrum/GsLocationFilter.tsx';
+import { GsTextInput } from '../genspectrum/GsTextInput.tsx';
 
 export type LocationFilterConfig = {
     locationFields: string[];
     initialLocation: LapisLocation;
     placeholderText: string;
+    label?: string;
 };
 
 export type DateRangeFilterConfig = {
-    initialDateRange: DateRangeOption;
+    defaultDateRange?: DateRangeOption;
+    initialDateRange?: DateRangeOption;
     dateRangeOptions: DateRangeOption[];
     earliestDate: string;
     dateColumn: string;
+    label?: string;
 };
+
+export type TextInputConfig = {
+    lapisField: string;
+    placeholderText?: string;
+    value?: string
+    label?: string;
+};
+
+export type BaselineFilterConfig =
+    | ({
+          type: 'date';
+      } & DateRangeFilterConfig)
+    | ({ type: 'text' } & TextInputConfig);
 
 export function BaselineSelector({
     locationFilterConfig,
     onLocationChange,
     dateRangeFilterConfig,
     onDateRangeChange,
+    baselineFilterConfigs,
+    onBaselineFilterConfigChange,
 }: {
     onLocationChange: (location: LapisLocation) => void;
     locationFilterConfig: LocationFilterConfig;
     onDateRangeChange: (dateRange: DateRangeOption) => void;
     dateRangeFilterConfig: DateRangeFilterConfig;
+    baselineFilterConfigs?: BaselineFilterConfig[];
+    onBaselineFilterConfigChange?: (baselineFilterConfigs: BaselineFilterConfig[]) => void;
 }) {
     return (
-        <div className='flex flex-col gap-2'>
-            <GsLocationFilter
-                fields={locationFilterConfig.locationFields}
-                onLocationChange={onLocationChange}
-                value={locationFilterConfig.initialLocation}
-                placeholderText={locationFilterConfig.placeholderText}
-            ></GsLocationFilter>
-            <GsDateRangeSelector
-                lapisDateField={dateRangeFilterConfig.dateColumn}
-                onDateRangeChange={onDateRangeChange}
-                earliestDate={dateRangeFilterConfig.earliestDate}
-                initialValue={dateRangeFilterConfig.initialDateRange}
-                dateRangeOptions={dateRangeFilterConfig.dateRangeOptions}
-            ></GsDateRangeSelector>
+        <div className={`flex flex-col gap-2`}>
+            <label className='form-control'>
+                <div className='label'>
+                    <span className='label-text'>
+                        {locationFilterConfig.label ?? locationFilterConfig.placeholderText}
+                    </span>
+                </div>
+                <GsLocationFilter
+                    fields={locationFilterConfig.locationFields}
+                    onLocationChange={onLocationChange}
+                    value={locationFilterConfig.initialLocation}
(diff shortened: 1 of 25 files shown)


website/src/styles/Expandable.tsx
JonasKellerer, line 1: This component is currently not used, since the overflow scroll does not work well with the selectors. I think I will delete it later.

website/src/views/View.ts
```diff
@@ -7,10 +7,13 @@ import { type PageStateHandler } from './pageStateHandlers/PageStateHandler.ts';
 import type { LineageFilterConfig } from '../components/pageStateSelectors/LineageFilterInput.tsx';
 import type { VariantFilterConfig } from '../components/pageStateSelectors/VariantFilterConfig.ts';
 import { type BreadcrumbElement } from '../layouts/Breadcrumbs.tsx';
+import type {BaselineFilterConfig} from "../components/pageStateSelectors/BaselineSelector.tsx";
 
 export type DatasetFilter = {
     location: LapisLocation;
     dateRange: DateRangeOption;
```
JonasKellerer, line 14: I think I can include this in the baselinefilters, but I wanted to keep it so there is less refactoring to be done after the PR of #486

website/src/components/pageStateSelectors/BaselineSelector.tsx
```diff
@@ -3,46 +3,144 @@ import type { DateRangeOption } from '@genspectrum/dashboard-components/util';
...
+                                                }
+
+                                                return conf.dateColumn === config.dateColumn
+                                                    ? { ...conf, initialDateRange: newDateRange }
+                                                    : conf;
+                                            });
+                                            if (onBaselineFilterConfigChange) {
+                                                onBaselineFilterConfigChange(newBaselineFilterConfigs);
+                                            }
+                                        }}
```
fengelniederhammer, line 105: IMO this looks more complicated than it needs to be. We could probably simplify this by having in some top level component:

```
const [dateFilters, setDateFilters] = useState(initialPageState.baseline.dateFilters);
const [textFilters, setTextFilters] = useState(initialPageState.baseline.textFilters);

   return {
    datasetFilter: {
      dateFilters,
      textFilters,
      ...
    },
    ...
  };
}, ...);
```

and passing the state down to this component:
```suggestion
                                        onDateRangeChange={(newDateRange) => {
                                            setDateFilters(previousDateFilters => ({
                                                ...previousDateFilters,
                                                [config.dateColumn]: newDateRange,
                                            }))
                                        }}
```

That would make the config actual config (because it doesn't change) and simplify handling state separately (reducing the overhead of dragging the config with us when updating the state).

fengelniederhammer: Essentially a lot of the simplification could also be achieved by turning `BaselineFilterConfig[]` into `{ [dateColumn]: BaselineFilterConfig }` (thus having a simple lookup instead of a loop, but leaving state and config in a single object).

website/src/views/pageStateHandlers/PageStateHandler.ts
```diff
@@ -16,10 +16,43 @@ export function toLapisFilterWithoutVariant(
...
+                    return {
+                        ...acc,
+                        [config.lapisField]: value,
+                        [config.lapisField]: value,
+                    };
+                }
+            }
+        },
+        {} as { [key: string]: string | undefined },
+    );
```
fengelniederhammer, line 49: Another benefit of separating the state from the config:
* You can have an array of configs -> The filter fields will be displayed in the same order as in the array (maybe important, I don't know).
* You can have a more structured "state", separating the fields by type in an object like `{ textFilters: ..., dateFilters: ... }`. That would turn this huge switch-case into two loops.

fengelniederhammer (commented):
Looks reasonable so far :+1: 

But I think I'm mostly convinced that we should separate state and config of the filters now (also see https://github.com/GenSpectrum/dashboards/pull/486#discussion_r1934053075)

website/src/views/View.ts
JonasKellerer: Now done.

website/src/components/pageStateSelectors/BaselineSelector.tsx
JonasKellerer: I now changed it, so that we have the config and the values separately.

fengelniederhammer (approved):
Looks good :+1:

website/src/views/pageStateHandlers/PageStateHandler.ts
```diff
@@ -35,18 +70,96 @@ export function toLapisFilterFromVariant(variantFilter: VariantFilter) {
...
+const columnsKey = 'columns';
 
 export function decodeFiltersFromSearch(search: URLSearchParams) {
     const filterMap = new Map<Id, Map<string, string>>();
 
+    const numFilters = Number.parseInt(search.get(columnsKey) ?? '0', 10);
+
+    for (let id = 0; id < numFilters; id++) {
+        filterMap.set(id, new Map<string, string>());
+    }
```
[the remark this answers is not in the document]
JonasKellerer: Could you elaborate on your comment here? I'm not sure what you are suggesting. Do you want to change lines 169-170 to the "continue"  solution?

website/tests/mainPage.spec.ts
```diff
@@ -32,9 +32,9 @@ test.describe('Main page', () => {
         await expect(page.getByRole('heading', { name: 'Inspect pathogen genomic data' })).toBeVisible();
     });
 
-    test(`should have a main organism navigation that links to the organism views`, async ({ page }) => {
-        for (const organism of organisms) {
-            for (const { linkName, title, expectedHeadline } of views) {
+    organisms.forEach((organism) => {
+        views.forEach(({ linkName, title, expectedHeadline }) => {
```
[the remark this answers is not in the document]
JonasKellerer: On my pc, there is no real time difference between the two versions. So I would keep it like this. But maybe you could test it as well? 
However, I ran into the timeout with the loop version and had to restart these tests.

website/src/views/pageStateHandlers/PageStateHandler.ts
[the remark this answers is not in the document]
fengelniederhammer: > Do you want to change lines 169-170 to the "continue" solution?

Yes. It's probably better because it will simply ignore filters with ids that actually don't occur.

website/tests/mainPage.spec.ts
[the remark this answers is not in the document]
fengelniederhammer: On GitHub, the tests took 4:37 min on the last run.

Two random builds from other branches took around 3:30. Let's try if we can save that minute by keeping at least one loop in the test.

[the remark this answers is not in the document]
JonasKellerer: I changed it, so we have only one loop.

PR #533, opened 2025-02-05
Some docs improvements

pflanze:
### Summary

Some random docs improvements while studying the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@
 
 The dashboards visualize genomic data obtained from several [LAPIS](https://github.com/GenSpectrum/LAPIS) instances.
 They utilize the [Dashboard Components](https://github.com/GenSpectrum/dashboard-components) library.
+
 Currently, we support the following organisms:
+
 - SARS-CoV-2
 - Influenza A
 - Influenza A/H5N1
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,8 +22,10 @@ when there is new data above a configurable threshold for certain variant of an
 ## Contents
 
 This monorepo contains the following packages:
-- [`website/`](./website): The dashboard website
-- [`backend/`](./backend): The backend for the dashboard website
+
+- [`website/`](./website): The dashboard website: delivery of the (basically static) HTML pages with the embedded web components ([Dashboard Components](https://github.com/GenSpectrum/dashboard-components), via npm).
+
+- [`backend/`](./backend): The backend for the dashboard website: the API from which the web components retrieve the data.
```
fengelniederhammer, line 28: That's not true. It retrieves the data from several LAPIS instances. They are unrelated to this project. The dashboards just gather the data.

The backend was written to support the notification features that is currently on hold. Either we leave it undocumented or we add a sentence like this.

website/src/pages/README.md
fengelniederhammer, line 1: IMO: We don't need this, since it's a standard Astro feature. I think we should rather advertise that we use Astro in a different place (the README maybe?) and point to the Astro docs. Reading the Astro docs should make clear how the code is structured.
We don't want to document the framework, only where we notable deviate from it.

(As I said, this is my opinion. I don't know how others think about this)

fengelniederhammer (changes requested):

website/src/pages/README.md
pflanze: I figured it's just helpful for someone who doesn't already know Astro. I've removed it now.

README.md
pflanze: I've pushed a fix, let me know what's still not correct.

pflanze: I've pushed another fix in 63b0e90e6701a48bea41b8f34f7b10cd7be9c2c6, hopefully this wording makes it clear what is retrieved from where, in a hopefully correct statement.

fengelniederhammer (approved):
Thanks for improving this!

In the long term, we can probably move docs to a separate place, but for now this is definitely better than before.

PR #535, opened 2025-02-07
Basic website docs

fengelniederhammer:
### Summary

I tried to summarize the basic concepts. I thought for now the CONTRIBUTING.md is actually a fitting place. If it grows too much, we can still move it somewhere else.


JonasKellerer (approved):
Nice. Good start. Now we need to keep the checkmark "Did you adapt the docs?" also for dashboards. (I am thinking of the filters section, which we need to adapt soon).

PR #537, opened 2025-02-09
Rename to filter

pflanze:
Updates the code to the change in https://github.com/GenSpectrum/dashboard-components/issues/709


pflanze:
Of course this requires the dependent PR to go in before the tests might run.

fengelniederhammer (approved):
I updated the components and squashed the commits.

pflanze:
I guess you had to go ahead because other work for components was coming in--co-releasing is time sensitive. And I was busy Tuesday and away Wednesday. Thanks for doing it! I checked what you did.

PR #545, opened 2025-02-10
CONTRIBUTING changes

pflanze:
### Summary

Some clarifications (hopefully correct?), and fixes. Also a last commit with open questions, we could drop that one and merge the rest, or finish that commit before merging.


CONTRIBUTING.md
```diff
@@ -23,9 +23,9 @@ We value clean code, here are some guidelines on what we consider clean code:
 
 The website shows dashboards for several organisms.
 The data sources are publicly available instances of [LAPIS](https://github.com/GenSpectrum/LAPIS) (one for each organism).
-Every organism has several views on the data.
+Every organism has several views on the data. "View" here means the whole page as displayed in the browser, each with their own base URL, which is mirroring the sub-paths of the `.astro` files in `website/src/pages/`.
 The different views of an organism highlight different aspects of the data,
-but they use the same LAPIS instance.
+but they use the same LAPIS instance (XX defined or configured where?).
```
fengelniederhammer, line 28: What do you mean by this? Where the LAPIS instances are defiend? Or where the dashboards have their config for which LAPIS they should access?

CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ The currently set filters are considered the "state" of the view.
 - The state is stored as query parameters in the URL,
   so that the user can bookmark the current view or send it to someone else.
 - Obviously, we must be able to reconstruct the state from a given URL.
-    - Upon a page reload, the values of the input fields in the filters section must be set according to the state in the URL.
+    - Upon a page reload, the values of the input fields in the filters section must be set according to the state in the URL. (XX how does this happen, is the state in a single location and pulled from there by the fields?)
```
fengelniederhammer, line 44: When we build the field (in React code), we have the state and know what piece of the state we need to access for that. But IMO that's too detailed and might change too easily to be documented here.

pflanze: What I have been, and for some things still am looking for, is a toplevel overview of what happens *and where*. The connection to the code is very helpful--it is really what I'm looking for. It's totally fine by me if this is just a hint here where to look, though, and then I would welcome to see the further details documented in the code itself.

I'm also happy to put documentation of the current state in a subfolder by date, "docs/old/2025-02-16/overview.md", with a "docs/old/README.md" that says "The docs in this folder are unmaintained. Use at your own risk.". That way nobody expects it to be current. It would still be easy to pick up again in the future and update if desired, and new contributors may still like reading through it.

fengelniederhammer: Would it be enough to give a pointer where you could start looking? We could say something like "start looking in src/pages/...", because I think this is very unlikely to change. From there, one should actually be able to navigate by itself.

pflanze: Yes, docs may be well splittable this way. But let me see how I end up collecting the knowledge--what works well for me to understand the material is to write it down, and in the first step I just have to write it down. Moving bits and pieces to their best place may have to happen in a second step.

pflanze:
Sorry, I wanted a rebase-then-create-merge-commit, now it made a plain merge (uglier).

PR #546, opened 2025-02-10
Docs work in progress

pflanze:
### Summary

This is documentation work I currently have pending; not ready to merge, just so you could see what I'm doing.

- The function documentation and types is probably useful, but has "XX" markers for open questions, and I'll probably have more of those documentation pieces over time.

- The components.md is just an attempt at getting an overview; it probably will never be useful to merge, at least not in this form.
-

### DIFF
--- a/website/docs/components.md
+++ b/website/docs/components.md
@@ -1,0 +1,315 @@
+# Components
+
+## React components
+
+* `src/components/ErrorReportInstruction.tsx`
+
+    - ErrorReportToastModal
+    - ErrorReportInstruction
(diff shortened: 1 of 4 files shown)


fhennig:
Hey! Just had a look at this since it's been floating around for a while.

Maybe we can get some clarification on the function comments from @fengelniederhammer or so, and then we remove the `components.md` (I agree, looks like it's difficult to make very useful, will be hard to keep updated) and then we can get it merged? 

Or we can move it out of 'in progress' if it's not urgent to finish this.

PR #554, opened 2025-02-12
chore(website): update components to 0.16

fengelniederhammer:
### Summary

This is breaking due to the changed default of the `height` attribute. We always set the height, so it doesn't affect us.


fengelniederhammer:
https://github.com/GenSpectrum/dashboard-components/pull/751 should fix the broken e2e tests.

PR #558, opened 2025-02-13
use aggregates bar chart

### LINKED ISSUE #476: Use Aggregate's bar chart
Cases where it would make sense to activate the new bar chart of the aggregate component that I see are:

- hosts
- For Pathoplexus organisms: Collection device, Collection method, Purpose of sampling, Sample type, Amplicon PCR primer scheme, Sequencing protocol -> Currently, there are only some data for West Nile where the values are known but hopefully there will be more Pathoplexus submissions soon
- Nextstrain clade for SARS-CoV-2 (needs yet to be added)
- Sub-lineages for mpox on the sequencing efforts page?

pflanze:
resolves #476 

### Summary

- add some return type definitions (useful for catching the missing fields for the other changes)

- GsAggregate astro component: require a `views` argument.

- Add `views` field to `additionalSequencingEffortsFields`, to enable
  configuration per such entry.

- Add the `views` argument statically to the remaining GsAggregate
  call sites.

### DIFF
--- a/website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
+++ b/website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
@@ -1,4 +1,6 @@
 ---
+import { views } from '@genspectrum/dashboard-components/util';
+
 import SelectVariant from './SelectVariant.astro';
 import { getDashboardsConfig } from '../../../config';
 import SingleVariantOrganismPageLayout from '../../../layouts/OrganismPage/SingleVariantOrganismPageLayout.astro';
@@ -116,17 +118,20 @@ const downloadLinks = noVariantSelected
                             title={subdivisionLabel}
                             fields={[subdivisionField]}
                             lapisFilter={variantLapisFilter}
+                            views={[views.table, views.bar]}
                         />
                     )}
                     <GsAggregate
                         title='Clades and lineages'
                         fields={getLineageFilterFields(view.organismConstants.lineageFilters)}
                         lapisFilter={variantLapisFilter}
+                        views={[views.table, views.bar]}
                     />
                     <GsAggregate
                         title='Hosts'
                         fields={[view.organismConstants.hostField]}
                         lapisFilter={variantLapisFilter}
+                        views={[views.table, views.bar]}
                     />
                 </ComponentsGrid>
                 <GsMutationsOverTime
--- a/website/src/components/genspectrum/GsAggregate.astro
+++ b/website/src/components/genspectrum/GsAggregate.astro
@@ -1,5 +1,5 @@
 ---
-import { type LapisFilter, views } from '@genspectrum/dashboard-components/util';
+import { type AggregateView, type LapisFilter } from '@genspectrum/dashboard-components/util';
 
 import { defaultTablePageSize } from '../../views/View';
 import ComponentWrapper from '../ComponentWrapper.astro';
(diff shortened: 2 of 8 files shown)


website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
```diff
@@ -116,17 +116,20 @@ const downloadLinks = noVariantSelected
...
                         title='Clades and lineages'
                         fields={getLineageFilterFields(view.organismConstants.lineageFilters)}
                         lapisFilter={variantLapisFilter}
+                        views={['bar', 'table']}
                     />
                     <GsAggregate
                         title='Hosts'
                         fields={[view.organismConstants.hostField]}
                         lapisFilter={variantLapisFilter}
+                        views={['bar', 'table']}
```
JonasKellerer, line 132: I'm not sure, if it would be better to have the default view the table view, since there are sometimes only very few hosts and therefore we have a large bar
```suggestion
                        views={['table', 'bar']}
```

grafik

The same applies to the other components.

JonasKellerer: I am more and more convinced to switch them arround. This is the view, where there is no "non-null" field value

grafik

JonasKellerer: What do you think @chaoran-chen @fengelniederhammer ?

JonasKellerer (approved):
LGTM. Only open question would be, what should be the default tab? I would favor the "table" tab, but lets see what the others say. Otherwise it is ready to merge.

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
```diff
@@ -116,17 +116,20 @@ const downloadLinks = noVariantSelected
                             title={subdivisionLabel}
                             fields={[subdivisionField]}
                             lapisFilter={variantLapisFilter}
+                            views={['bar', 'table']}
```
fengelniederhammer, line 119: I would prefer to also use `views.table` for those values here.

pflanze: You mean show *only* the table, right? (That is e.g. the country, and the only entry in the table, hence bar chart makes no sense, right?)

fengelniederhammer: No, but use `views.something` instead of hardcoded strings.

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
fengelniederhammer: I'm not sure which one is better. It depends on the diagram and the organism... I don't have a strong preference.

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
pflanze: Ah, I see. Interestingly here it didn't need 'as const'.

fengelniederhammer: It didn't need `as const` here, because you directly assign `'table'` to an argument that has type `'table' | 'bar'`.

In the other case you stored it in a variable that didn't have an explicit type, so Typescript inferred the type of `'table'` as `string`. Then you tried to assign that to a field that needed the explicit type `'table'`.

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
chaoran-chen: Hmm, I see the issue. Okay, happy to switch it around and have the table as the default view.

PR #569, opened 2025-02-18
feat(website): change default date range to all times for Ebola and CCHF

fengelniederhammer:
### Summary

because "last year" shows no data


pflanze (approved):
Looks good to me.

PR #571, opened 2025-02-19
feat(website): add labels to variant filter inputs

pflanze:
### Summary

Without labels, it can be unclear what the entry is about, and for consistency with the other entries.

### DIFF
--- a/website/src/components/genspectrum/GsMutationFilter.tsx
+++ b/website/src/components/genspectrum/GsMutationFilter.tsx
@@ -37,10 +37,15 @@ export function GsMutationFilter({
     }, [onMutationChange]);
 
     return (
-        <gs-mutation-filter
-            width={width}
-            initialValue={JSON.stringify(initialValue)}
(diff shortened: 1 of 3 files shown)


website/src/components/genspectrum/GsMutationFilter.tsx
```diff
@@ -37,10 +37,15 @@ export function GsMutationFilter({
...
 
     return (
-        <gs-mutation-filter
-            width={width}
-            initialValue={JSON.stringify(initialValue)}
-            ref={mutationFilterRef}
-        ></gs-mutation-filter>
+        <label className='form-control'>
+            <div className='label'>
+                <span className='label-text'>Mutation</span>
```
fengelniederhammer, line 42: I think I would make it plural:
```suggestion
                <span className='label-text'>Mutations</span>
```

fengelniederhammer (approved):
LGTM

PR #572, opened 2025-02-19
feat(website): add quickstart links to single variant pages when no variant filter is set

### LINKED ISSUE #565: Basic guidance: 3 suggestions on single variant pages

fengelniederhammer:
resolves #565

### Summary

I don't really know which variants are best to put there, so I just chose something that had decent numbers. 

### Screenshot

Covid:
image

RSV-A:
image


fengelniederhammer:
I changed the layout as discussed: When no variant selected then we still show the help text, but in a box and we also show the graphs, even though they don't show "variant data". This way users can already see data that helps them choose a lineage/mutation that they want to investigate.

H5N1:
image

Covid:
image

website/src/components/views/analyzeSingleVariant/CollectionsList.tsx
```diff
@@ -9,6 +9,8 @@ import { Routing } from '../../../views/routing.ts';
 import { useErrorToast } from '../../ErrorReportInstruction.tsx';
 import { withQueryProvider } from '../../subscriptions/backendApi/withQueryProvider.tsx';
 
+export const collectionVariantClass = 'border bg-white px-4 py-2 hover:bg-cyan';
```
JonasKellerer, line 12:
```suggestion
export const collectionVariantClassName = 'border bg-white px-4 py-2 hover:bg-cyan';
```
Nitpick.

website/src/components/views/analyzeSingleVariant/QuickstartLinks.astro
```diff
@@ -0,0 +1,153 @@
...
+import { singleVariantViewKey } from '../../../views/viewKeys';
+
+type SingleVariantOrganism = OrganismWithViewKey<typeof singleVariantViewKey>;
+interface Props {
+    view: ViewsMap[SingleVariantOrganism][typeof singleVariantViewKey];
+}
+
+const { view } = Astro.props;
+
+const predefinedVariants: Record<SingleVariantOrganism, VariantFilter[]> = {
```
JonasKellerer, line 16: Should we put this into the view config (meaning `WestNileConstants` for example)? I think this is currently one central point for all the config.

website/src/components/views/analyzeSingleVariant/SelectVariant.astro
```diff
@@ -2,15 +2,14 @@
 import { PageHeadline } from '../../../styles/containers/PageHeadline';
 ---
 
-<div class='mt-20 flex max-w-xl flex-col'>
-    <div class='mx-auto'>
-        <PageHeadline>Analyze a single variant</PageHeadline>
-    </div>
-    <div class='mx-2 flex max-w-xl flex-col gap-2 text-center'>
+<div class='flex w-full flex-col items-center justify-center rounded border-2 border-sand bg-sandMuted p-8'>
```
JonasKellerer, line 5: Its only my preference, but the sand is a bit too much for me. I would use primary, since this should foucs the attention of the user. It is no real error, that no variant was inserted.

```suggestion
<div class='flex w-full flex-col items-center justify-center rounded border-2 border-primary p-8'>
```

JonasKellerer (approved):
I like this design. One can use the page, but it is still clear that the user needs to take an action. 
Not this PR, but I think we should improve the Dropdown/Select of the variants. The covid one is quiet long and takes a lot of the screen. And the others seem a bit barebones.

PR #574, opened 2025-02-19
refactor(website): make the collection button links

fengelniederhammer:
### Summary

so that the browser can tell the user where they link to


JonasKellerer (approved):
Good idea.

PR #583, opened 2025-02-25
feat(website): use auto height for tables and mutation over time

### LINKED ISSUE #566: Use auto height for mutations over time

### LINKED ISSUE #567: Try auto height for tables

JonasKellerer:
Resolves #567
Resolves #566

### Summary

Uses the new auto height feature for tables and mutations over time component. 

As discussed: We have still some whitespace below the tables, when there are too few entries (see also screenshot). We could try to use two columns instead of the grid. However, this would undermine the horizontal aligments of the tables. Therefore we keep it like this.

### Screenshot

grafik

### DIFF
--- a/website/src/components/genspectrum/GsAggregate.astro
+++ b/website/src/components/genspectrum/GsAggregate.astro
@@ -6,13 +6,14 @@ import ComponentWrapper from '../ComponentWrapper.astro';
 
 interface Props {
     title: string;
-    height?: string;
     fields: string[];
     lapisFilter: LapisFilter;
     views: AggregateView[];
+    height?: string;
+    pageSize?: number;
 }
 
-const { title, height, fields, lapisFilter, views } = Astro.props;
+const { title, height, fields, lapisFilter, views, pageSize } = Astro.props;
 ---
 
 <ComponentWrapper title={title} height={height}>
@@ -20,7 +21,7 @@ const { title, height, fields, lapisFilter, views } = Astro.props;
         views={JSON.stringify(views)}
         fields={JSON.stringify(fields)}
         lapisFilter={JSON.stringify(lapisFilter)}
-        pageSize={defaultTablePageSize}
+        pageSize={pageSize ?? defaultTablePageSize}
         width='100%'
-        height='100%'></gs-aggregate>
+        height={height ? '100%' : undefined}></gs-aggregate>
 </ComponentWrapper>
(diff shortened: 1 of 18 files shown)


website/src/components/genspectrum/GsAggregate.astro
```diff
@@ -6,21 +6,22 @@ import ComponentWrapper from '../ComponentWrapper.astro';
...
 <ComponentWrapper title={title} height={height}>
     <gs-aggregate
         views={JSON.stringify(views)}
         fields={JSON.stringify(fields)}
         lapisFilter={JSON.stringify(lapisFilter)}
-        pageSize={defaultTablePageSize}
+        pageSize={pageSize ?? defaultTablePageSize}
         width='100%'
-        height='100%'></gs-aggregate>
+        height={height ? '100%' : ''}></gs-aggregate>
```
fengelniederhammer, line 26: I think this also works:
```suggestion
        height={height ? '100%' : undefined}></gs-aggregate>
```

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
```diff
@@ -107,16 +108,21 @@ const downloadLinks = noVariantSelected
...
                 baselineLapisFilter={datasetLapisFilter}
                 sequenceType='nucleotide'
+                pageSize={8}
+                height={ComponentHeight.medium}
             />
             <GsMutations
                 lapisFilter={variantLapisFilter}
                 baselineLapisFilter={datasetLapisFilter}
                 sequenceType='amino acid'
+                pageSize={10}
```
fengelniederhammer, line 125: Why are AA and nuc mutations different? I would expect that they have the same height and page size?

website/src/components/views/sequencingEfforts/GenericSequencingEffortsPage.astro
```diff
@@ -92,15 +94,16 @@ const lineageFilterFields = getLineageFilterFields(view.organismConstants.lineag
...
                 />
             )
         }
         <GsAggregate
             title='Hosts'
-            height={ComponentHeight.large}
             fields={[view.organismConstants.hostField]}
             lapisFilter={lapisFilter}
             views={[views.table, views.bar]}
+            pageSize={10}
```
fengelniederhammer, line 106: Nitpick: It's a bid weird that the page sizes are not equal here:
image

fengelniederhammer (approved):
Looks a lot better than before IMO. A lot less scrollable containers :+1:

website/src/components/views/analyzeSingleVariant/GenericAnalyzeSingleVariantPage.astro
JonasKellerer: I tried to have the same height as the GsPrevalenceOverTime component on the two row view. However, as you mentioned below, when we have three rows this does not work any longer. So I changed it, that the component uses a pageSize of 10.
grafik

website/src/components/views/sequencingEfforts/GenericSequencingEffortsPage.astro
JonasKellerer: See comment above. It now uses the auto height for most of the plots.
grafik

PR #585, opened 2025-02-25
refactor(website): merge ExtendedConstants into OrganismConstants

fengelniederhammer:
### Summary

The original idea was to only have a minimal set of value in OrganismConstants and extend when necessary. Reality showed that we never use a minimal set, but always the full set instead. The separation became irrelevant, so let's remove it.


JonasKellerer (approved):
LGTM

PR #589, opened 2025-02-27
feat(website): add host as a filter option

anna-parker:
resolves #

### Summary

image

### DIFF
--- a/website/src/views/cchf.ts
+++ b/website/src/views/cchf.ts
@@ -40,6 +40,12 @@ class CchfConstants implements ExtendedConstants {
             dateColumn: 'sampleCollectionDateRangeLower',
             label: 'Sample collection date',
         },
+        {
+            lapisField: 'hostNameScientific',
+            placeholderText: 'Host',
+            type: 'text' as const,
+            label: 'Host',
+        },
     ];
     public readonly mainDateField: string;
     public readonly locationFields: string[];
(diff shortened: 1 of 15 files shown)


website/src/views/cchf.ts
```diff
@@ -40,6 +40,12 @@ class CchfConstants implements ExtendedConstants {
             dateColumn: 'sampleCollectionDateRangeLower',
             label: 'Sample collection date',
         },
+        {
+            lapisField: 'hostNameScientific',
```
fengelniederhammer, line 44: We also have `public readonly hostField: string;` which is filled from the config (and I believe the values are the same). IMO it would be fine to remove it from the config and simply hardcode it in this file here.

Would you like to do that? I think it would be good to avoid the duplication. But we can also do that afterwards.

I imagine that it could look similar to what we do for `earliestDate`.

Pinging @JonasKellerer to make sure he's also ok with that.

JonasKellerer: I would be fine with both, ith leaving it in, or removing it. But having a constant like earliest data would be nice, to indicate that we mean the same field.

anna-parker: I created an issue: https://github.com/GenSpectrum/dashboards/issues/593

fengelniederhammer (approved):
Thanks :)

PR #600, opened 2025-03-03
fix(website): remember the collection id on covid single variant page when changing filters

### LINKED ISSUE #573: Covid single variant page: changing the date makes the page forget the selected collection
* go to covid single variant page
* change the collection, e.g. to `#2`
* select a collection
* -> The page reloads and display the data of the collection. The collection selector still shows the selection `#2`.
* Change the date (e.g. to "all times")
* The collection selector shows the selection `#1`.

IMO it should remember that the user selected a different collection.

fengelniederhammer:
resolves #573


website/src/components/subscriptions/backendApi/backendService.spec.ts
```diff
@@ -161,12 +161,6 @@ describe('backendService', () => {
         await expect(backendService.deleteSubscription({ subscriptionId, userId })).resolves.to.deep.equal('');
     });
 
-    test('should throw error when backend is not reachable', async () => {
-        await expect(backendService.getSubscriptions({ userId: '123' })).rejects.to.deep.equal(
-            new BackendNotAvailable(DUMMY_BACKEND_URL),
-        );
-    });
```
fengelniederhammer, line 168: I didn't know how to implement this test with explicitly defining a mocked request. Without one, it would fail, because it issued unhandled requests.

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.spec.tsx
fengelniederhammer, line 1: This is by far not perfect yet, but it's a first test that somehow uses the web components in a unit test. I'd suggest to improve from here and check whether we can have more tests based on this setup.

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.tsx
```diff
@@ -1,43 +1,49 @@
...
 }) {
-    const [datasetFilterState, setDatasetFilterState] = useState(datasetFilter);
+    const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
+    const variantFilterConfig = useMemo(() => {
+        return getVariantFilterConfig(
+            view.organismConstants.lineageFilters,
+            pageState.variantFilter,
+            view.organismConstants.useAdvancedQuery,
+        );
+    }, []);
```
JonasKellerer, line 32: I'm not sure, why the linter does not pick up on this, but we should add the pageState and the view as dependencies, don't we?

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.tsx
```diff
@@ -1,43 +1,49 @@
...
+    const [datasetFilterState, setDatasetFilterState] = useState(pageState.datasetFilter);
 
     const [variantFilterConfigState, setVariantFilterConfigState] = useState<VariantFilterConfig>(variantFilterConfig);
 
-    const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
+    const baselineFilterConfigs = view.organismConstants.baselineFilterConfigs;
 
     const newPageState = useMemo(() => {
         return {
+            ...pageState,
```
JonasKellerer, line 42: When I see this correctly, we use the pageState here, so we still have the `{ collectionId?: number }` in the page state. When only looking at the type, one would think that the use of the `...pageState` is not necessary. Should we somehow capture this in the type, or do we argue, that the Selector does not need to have the type, because we don't need the collectionId inside the selector?

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.spec.tsx
JonasKellerer: Good start and then lets see.

website/src/components/subscriptions/backendApi/backendService.spec.ts
JonasKellerer: I had the same issue. Since we don't need the backend for now, I think it is fine.

website/vitest.setup.ts
```diff
@@ -148,3 +159,260 @@ beforeAll(() => testServer.listen({ onUnhandledRequest: 'warn' }));
 afterAll(() => testServer.close());
 
 afterEach(() => testServer.resetHandlers());
+
+export const testOrganismsConfig = {
```
JonasKellerer, line 163: I hope this becomes easier, when we've removed the config and put it into the files.

JonasKellerer (approved):
LGTM

website/vitest.setup.ts
fengelniederhammer: I guess so. Copying the whole config to the tests isn't ideal, but I couldn't think of a better alternative for now.

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.tsx
fengelniederhammer: I'll check in a follow up PR.

website/src/components/pageStateSelectors/SingleVariantPageStateSelector.tsx
fengelniederhammer: I added a comment instead. I'm not sure whether adding something to the type would make it better.

PR #606, opened 2025-03-04
refactor(website): separate variant filter config and variant filter state

fengelniederhammer:
### Summary

Main changes:
* make the pageState a state of the state selector so that they don't have to be computed.
  Instead, they should be the state that everything else is computed from.
* "filter configs" are "static", i.e. don't depend on the state.


website/src/views/pageStateHandlers/PageStateHandler.ts
```diff
@@ -181,11 +181,16 @@ export function encodeMultipleFiltersToUrlSearchParam(filters: Map<Id, Map<strin
...
 
-    for (const [id, filter] of filters) {
-        for (const [key, value] of filter) {
-            search.append(`${key}${variantFilterUrlDelimiter}${id}`, value);
-        }
-    }
+    filters
+        .entries()
+        .toArray()
+        .toSorted(([id1], [id2]) => id1 - id2)
```
JonasKellerer, line 187: Just out of curiosity: Why do we sort them here?

website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
```diff
@@ -1,53 +1,56 @@
...
     const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
+    const variantFilterConfig = useMemo(
+        () => ({
+            lineageFilterConfigs: view.organismConstants.lineageFilters,
+            mutationFilterConfig: { enabled: true },
+            variantQueryConfig: { enabled: view.organismConstants.useAdvancedQuery },
+        }),
+        [view.organismConstants],
+    );
+    const [currentPageState, setCurrentPageState] = useState(pageState);
```
JonasKellerer, line 35: Just an idea: maybe we want to name the pageState we get as a prop: `initialPageState` and then have this as `pageState`.

website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
```diff
@@ -1,53 +1,56 @@
...
     locationFilterConfig,
-    variantFilterConfig,
     filterId,
     pageState,
     organismViewKey,
     organismsConfig,
-    baselineFilterConfigs,
-    datasetFilter,
 }: {
     locationFilterConfig: LocationFilterConfig;
```
JonasKellerer, line 20: You have now moved the config of the baseline and variant filter inside the component and read it from the view. We could also just provide the view and use this for all other things. We already know the view on the consuming side. This would then also get rid of the `organismsConfig`, `organismViewKey` and `locationFilterConfig`. This would make it also easier to test, since we can provide the view directly and dont have to provide the config. 
I think the same is true for all the other PageSelectors.

JonasKellerer: Maybe not this PR, but we could also get rid of the locationFilterConfig and move it inside.

website/src/components/pageStateSelectors/CompareVariantsPageStateSelector.tsx
```diff
@@ -1,56 +1,45 @@
...
-            }),
+    const variantFilterConfigs = useMemo(() => {
+        return new Map(
+            currentPageState.variants.entries().map(([id]) => [
+                id,
+                {
+                    lineageFilterConfigs: view.organismConstants.lineageFilters,
+                    mutationFilterConfig: { enabled: true },
+                    variantQueryConfig: { enabled: view.organismConstants.useAdvancedQuery },
+                },
```
JonasKellerer, line 35: This is constant and is used in all our StateSelectors. I would be in favor of putting this as a function, or as const in the view like the baselineFilterConfigs.

website/src/components/pageStateSelectors/VariantSelector.tsx
```diff
@@ -1,87 +1,94 @@
...
+                        type='checkbox'
+                        className='checkbox checkbox-xs'
+                        onChange={() => setIsInVariantQueryMode((value) => !value)}
+                        checked={isInVariantQueryMode}
+                    />
+                </label>
+            )}
 
-            <div className={variantFilterConfig.isInVariantQueryMode ? '' : 'hidden'}>
+            <div className={isInVariantQueryMode ? '' : 'hidden'}>
```
JonasKellerer, line 49:
```suggestion
{isInVariantQueryMode ? <div> ... Normal Filter </div> : <div>... variant Query </div>
```
We could make it clear, that either one of these filters is used.

website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
JonasKellerer: Also for the other selectors.

JonasKellerer (changes requested):
Nice. Good to have this now separated. Thank you. On some places, I would go a bit further.

website/src/views/pageStateHandlers/PageStateHandler.ts
fengelniederhammer: Just to be sure. I changed this piece of code such that:
* The page state itself has stable ids that may have gaps, if you delete a variant in between.
* The URL always has ids starting from 0 (this code "squashes" them).

I thought sorting doesn't hurt to make sure that at least the order of variants is preserved.

JonasKellerer: Thanks.

website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
JonasKellerer: I see. Thank you. I always forget the functions. Then lets add a ticket for the location filter. :)

fengelniederhammer: -> #618

website/src/components/pageStateSelectors/SequencingEffortsPageStateSelector.tsx
```diff
@@ -23,12 +23,12 @@ export function SequencingEffortsPageStateSelector({
...
     const variantFilterConfig = useMemo(
-        () => ({
-            lineageFilterConfigs: view.organismConstants.lineageFilters,
-            mutationFilterConfig: { enabled: false },
-            variantQueryConfig: { enabled: false },
-        }),
-        [view.organismConstants.lineageFilters],
+        () =>
+            makeVariantFilterConfig(view.organismConstants, {
+                enableMutationFilter: true,
```
JonasKellerer, line 28: Is this change from enabled false to true intended?

website/src/components/pageStateSelectors/VariantSelector.tsx
```diff
@@ -17,6 +18,24 @@ export type VariantFilterConfig = {
     };
 };
 
+type Options = {
+    enableMutationFilter: boolean;
```
JonasKellerer, line 22: In this PR, we have no place, where we want to set it to false. Do you think we need this in the futur?

JonasKellerer (approved):
Nice. LGTM.

website/src/components/pageStateSelectors/VariantSelector.tsx
JonasKellerer: Ah, just saw my comment above. Maybe this is the place. :)

PR #611, opened 2025-03-06
feat(website): add 'Sub-lineages Neststrain Clade' for Sars-Cov-2

### LINKED ISSUE #559: Nextstrain clade for SARS-CoV-2
(Originally mentioned in #476)

pflanze:
The reason to implement it using a separate component instantiation rather than adding another `LineageFilterConfig` to `CovidConstants.lineageFilters` (which adds another column to the existing output) is that the ~corresponding Nextstrain Clade for a Nextclade pango lineage is not actually associated with the same data, thus it would be wrong. Also, the bar chart currently only shows the first column, thus would not show the Nextstrain Clade names.

resolves #559

### DIFF
--- a/website/src/pages/covid/single-variant.astro
+++ b/website/src/pages/covid/single-variant.astro
@@ -127,8 +127,15 @@ const organismConfig = getDashboardsConfig().dashboards.organisms;
                 pageSize={10}
             />
             <GsAggregate
-                title='Sub-lineages'
+                title='Nextclade pango lineage'
                 fields={getLineageFilterFields(view.organismConstants.lineageFilters)}
+                lapisFilter={variantFilter}
+                views={[views.table, views.bar]}
+                pageSize={10}
+            />
+            <GsAggregate
+                title='Nextstrain Clade'
+                fields={getLineageFilterFields(view.organismConstants.nextstrainCladeLineageFilters)}
                 lapisFilter={variantFilter}
                 views={[views.table, views.bar]}
                 pageSize={10}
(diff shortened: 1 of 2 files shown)


website/src/pages/covid/single-variant.astro
```diff
@@ -133,6 +133,13 @@ const organismConfig = getDashboardsConfig().dashboards.organisms;
                 views={[views.table, views.bar]}
                 pageSize={10}
             />
+            <GsAggregate
+                title='Sub-lineages Nextstrain Clade'
```
fengelniederhammer, line 137:
```suggestion
                title='Nextstrain Clade'
```

IMO we could also change the title of the other diagram from `Sub-lineage` to `Nextstrain Pango Lineage`

website/src/views/covid.ts
```diff
@@ -45,6 +45,15 @@ class CovidConstants implements OrganismConstants {
...
     ];
+    // Same as `lineageFilters` but for Nextstrain Clade, used for 'Sub-lineages Nextstrain Clade'
+    public readonly nextstrainCladeLineageFilters: LineageFilterConfig[] = [
+        {
+            lapisField: 'nextstrainClade',
+            placeholderText: 'Nextstrain clade',
+            filterType: 'lineage' as const,
+            initialValue: undefined,
+        },
+    ];
```
fengelniederhammer, line 56: IMO we could also inline that as it's used in a single place. I don't see yet whether we would reuse it somewhere else (and if so, I don't know how we would reuse it).

But it's also fine to leave it here for now.

fengelniederhammer (approved):
LGTM

website/src/views/covid.ts
pflanze: I think there's some value in keeping the abstraction the same / keeping the data in the same abstraction level as for the other diagram.

PR #615, opened 2025-03-10
feat(website): add influenza B and B/Victoria dashboards

anna-parker:
resolves https://github.com/GenSpectrum/dashboards/issues/599

### Summary

Adds influenza b and b-victoria dashboards
https://github.com/GenSpectrum/dashboards/pull/596 should be merged first.

### Screenshot

image

PR #624, opened 2025-03-13
feat(website): improve overview for influenza A and B

anna-parker:
resolves #

### Summary

Link to influenza A subtypes and influenza B lineages on the overview page. 

### Screenshot

image


website/src/pages/flu/[...slug].astro
Comment (fengelniederhammer) [inline website/src/pages/flu/[...slug].astro:1]: This seems to be a leftover from rebasing?

website/src/pages/influenza-a/index.astro
```diff
@@ -1,6 +1,9 @@
 ---
-import GenericOverview from '../../components/views/overview/GenericOverview.astro';
+import LinkedOverview from '../../components/views/overview/LinkedOverview.astro';
+import { getPathogenMegaMenuSections } from '../../layouts/base/header/getPathogenMegaMenuSections';
 import { Organisms } from '../../types/Organism';
+
+const subtypes: (keyof ReturnType<typeof getPathogenMegaMenuSections>)[] = ['h5n1', 'h3n2', 'h1n1pdm'];
```
fengelniederhammer, line 6: Let's please use the constants from `Organisms`. (Also I think you won't need explicit type annotations then, because TS will recognize them as const values. Nitpick: IMO we could also inline `subtypes`).

website/src/pages/influenza-b/index.astro
```diff
@@ -1,6 +1,13 @@
 ---
-import GenericOverview from '../../components/views/overview/GenericOverview.astro';
+import LinkedOverview from '../../components/views/overview/LinkedOverview.astro';
+import { getPathogenMegaMenuSections } from '../../layouts/base/header/getPathogenMegaMenuSections';
 import { Organisms } from '../../types/Organism';
+
+const lineages: (keyof ReturnType<typeof getPathogenMegaMenuSections>)[] = ['victoria'];
```
fengelniederhammer, line 6: Let's please use the constants from `Organisms`. (Also I think you won't need explicit type annotations then, because TS will recognize them as const values. Nitpick: IMO we could also inline `lineages`).

website/src/components/views/overview/LinkedOverview.astro
```diff
@@ -0,0 +1,80 @@
...
+                                <span class={`btn btn-soft ${section.headlineBackgroundColor}`}>
+                                    Go to {entry.label}
+                                </span>
+                            </div>
+                        </a>
+                    </div>
+                );
+            })
+        }
+    </div>
```
fengelniederhammer, line 49: This is copied from the `GenericOverview`, isn't it? I'm thinking whether we can reuse that?

I think it would be best to introduce a `<slot/>` in the `GenericOverview`, keep using the `GenericOverview` in the index.astros and render a common component into that slot? Do you have other suggestions?

website/src/components/views/overview/LinkedOverview.astro
```diff
@@ -0,0 +1,80 @@
...
+                                    menuListEntryDecoration={entry.underlineColor}
+                                    href={entry.href}
+                                    external={entry.externalLink}
+                                >
+                                    {entry.label}
+                                </MainMenuItem>
+                            ))}
+                        </ul>
+                    </div>
+                </div>
```
fengelniederhammer, line 76: Is this any different from what we use on the main page? Can we extract it into a component and reuse it here?

website/src/components/views/overview/LinkedOverview.astro
```diff
@@ -0,0 +1,80 @@
...
+                        </a>
+                    </div>
+                );
+            })
+        }
+    </div>
+    <h1 class='mt-10 mb-4 text-xl font-bold'>{linkedOrganismsTitle}</h1>
+    <div class='items mx-auto mb-8 flex max-w-[1200px] flex-wrap justify-center gap-x-4 gap-y-8'>
+        {
+            Object.values(subsections).map((section) => (
```
fengelniederhammer, line 53: Do we really need the `Object.values`?
```suggestion
            subsections.map((section) => (
```

website/src/components/views/overview/LinkedOverview.astro
```diff
@@ -0,0 +1,80 @@
...
+    organism: keyof ReturnType<typeof getPathogenMegaMenuSections>;
+    linkedOrganismsTitle: string;
+    linkedOrganisms: (keyof ReturnType<typeof getPathogenMegaMenuSections>)[];
+};
+
+const { organism, linkedOrganismsTitle, linkedOrganisms } = Astro.props;
+
+const section = getPathogenMegaMenuSections()[organism];
+
+const allSections = getPathogenMegaMenuSections();
```
fengelniederhammer, line 19:
```suggestion
const allSections = getPathogenMegaMenuSections();
const section = allSections[organism];
```

fengelniederhammer (changes requested):
Nice idea, looks good!

Let's just clean up the code a little.

website/src/pages/flu/[...slug].astro
Comment (anna-parker) [reply website/src/pages/flu/[...slug].astro:1]: oh wupps I accidentally moved this in https://github.com/GenSpectrum/dashboards/pull/615 but this now is correct.

website/src/pages/influenza-a/index.astro
```diff
@@ -1,6 +1,8 @@
 ---
-import GenericOverview from '../../components/views/overview/GenericOverview.astro';
-import { Organisms } from '../../types/Organism';
+import LinkedOverview from '../../components/views/overview/LinkedOverview.astro';
+import { Organisms, type Organism } from '../../types/Organism';
+
+const subtypes: Organism[] = ['h5n1', 'h3n2', 'h1n1pdm'];
```
fengelniederhammer, line 5:
```suggestion
const subtypes = [Organisms.h5n1, Organisms.h3n2, Organisms.h1n1pdm];
```

website/src/pages/influenza-b/index.astro
```diff
@@ -1,6 +1,12 @@
 ---
-import GenericOverview from '../../components/views/overview/GenericOverview.astro';
-import { Organisms } from '../../types/Organism';
+import LinkedOverview from '../../components/views/overview/LinkedOverview.astro';
+import { Organisms, type Organism } from '../../types/Organism';
+
+const lineages: Organism[] = ['victoria'];
```
fengelniederhammer, line 5:
```suggestion
const lineages = [Organisms.victoria];
```

website/src/components/MainMenu.astro
```diff
@@ -0,0 +1,40 @@
...
+                                {entry.label}
+                            </MainMenuItem>
+                        ))}
+                    </ul>
+                </div>
+            </div>
+        ))
+    }
+</div>
+---
```
fengelniederhammer, line 40:
```suggestion
```

fengelniederhammer (approved):
Nice :)

website/src/pages/influenza-a/index.astro
anna-parker: Ah thats what you meant! Thanks!

Issue #581, opened 2025-02-25
Gene plot


chaoran-chen:
As a user, I would like to be able to easily find out to which gene(s) a mutation belongs.

### Examples from other projects

A common approach is to show a plot that visualizes the genes. Outbreak.info shows their list of mutations this way:

![Image](https://github.com/user-attachments/assets/4cb6517b-6a88-45d6-8c67-04125654bfdd)

Nextstrain also shows the genes below their nucleotide entropy plot:

![Image](https://github.com/user-attachments/assets/002947b5-376a-43d4-b35c-56365e4d7f4a)

A Python library with more examples of such plots is here: https://edinburgh-genome-foundry.github.io/DnaFeaturesViewer/

### Idea

@anna-parker and I discussed that it would be useful to have such a plot on our dashboards as well. However, the challenge is to find a good location to place it because we have mutation lists in many components. Our idea is to have a button to show/hide such a plot fixed in a corner:

![Image](https://github.com/user-attachments/assets/9acd44f5-c72c-4ea3-ac7b-926f25323121)

When visible and when the user hovers over a mutation, it will show a line to indicate the corresponding location on the plot:

![Image](https://github.com/user-attachments/assets/9889c81a-b308-45d4-80ff-c524bcb9a8ec)

PR #628, opened 2025-03-17
feat(website): add genome data viewer

anna-parker:
partially resolves https://github.com/GenSpectrum/dashboards/issues/581

### Summary

This uses the genome-data-viewer component from https://github.com/GenSpectrum/dashboard-components/pull/814 in a later PR I will add an info text for individual genes.


website/src/components/views/overview/GenericOverview.astro
```diff
@@ -42,5 +50,34 @@ const breadcrumbs = [...defaultBreadcrumbs, ...paths[organism].breadcrumbs];
...
+                            <span slot='infoText'>
+                                <slot name='infoText' />
+                            </span>
+                        </gs-genome-data-viewer>
+                    </div>
+                ))
+            }
+        </div>
+    </ComponentWrapper>}
+    </div>
```
JonasKellerer, line 81: I think I would put this into the 'post-navigation' slot. This way, we do not have to add, if there should be a genome view or not.

Then put this into an own astro component.
And then use this slot in each index page (or not if not needed).
Example for cchf
```
---
import GenericOverview from '../../components/views/overview/GenericOverview.astro';
import { Organisms } from '../../types/Organism';
import GenomeView from "../../components/views/overview/GenomeView.astro";
---

<GenericOverview organism={Organisms.cchf}>
    <GenomeView slot="post-navigation" organism={Organisms.cchf} lengthScaleFactor={2} />
</GenericOverview>
```

JonasKellerer: I also think it would be great that we could have the information of the scale factor in the `OrganismConstants` and don't have to provide it here. However, it is currently not possible to get the organsim constants directly from the organism name (as far as I know). So this would be maybe a task for the future.

It would then also get rid of the funny `getPathogenMegaMenuSections` and the `getGenomeDataSections` and get them directly from each organismConfig. But this is not part of this PR.

website/src/layouts/base/header/getGenomeData.ts
JonasKellerer, line 1: I would not put this into the `layouts` folder, since this has nothing to do with the layout. I would put it next to the new astro component for the viewer. I would then put the astro component, next to the `GenericOverview.astro` file.

website/src/types/Organism.ts
```diff
@@ -28,6 +28,13 @@ export const organismConfig = {
         backgroundColor: 'bg-sandMuted',
         backgroundColorFocus: 'group-hover:bg-sand',
         borderEntryDecoration: 'hover:border-sand',
+        genome: [
+            {
+                gff3Source:
+                    'https://raw.githubusercontent.com/nextstrain/nextclade_data/refs/heads/master/data/nextstrain/sars-cov-2/BA.2/genome_annotation.gff3',
+                genomeLength: 29903, // check this
```
JonasKellerer, line 35: Is this still to be checked?

website/src/layouts/base/header/getGenomeData.ts
```diff
@@ -0,0 +1,27 @@
+import { type Organism, organismConfig } from '../../../types/Organism.ts';
+
+export type GenomeData = {
+    url: string;
```
JonasKellerer, line 4: In the genome we call this `gff3Source`. I think it would be nice to have the same naming here.

website/src/components/views/overview/GenericOverview.astro
```diff
@@ -1,24 +1,32 @@
...
 import { iconMapping } from '../../iconCss';
 
 type Props = {
     organism: keyof ReturnType<typeof getPathogenMegaMenuSections>;
+    addGenomeDataViewer?: boolean;
+    lengthScaleFactor?: number;
 };
 
-const { organism } = Astro.props;
+const { organism, addGenomeDataViewer = true, lengthScaleFactor } = Astro.props;
```
JonasKellerer, line 17: In the new file, I would add a default value here, instead of line 20.
```suggestion
const { organism, addGenomeDataViewer = true, lengthScaleFactor = 1 } = Astro.props;
```

website/src/components/views/overview/GenericOverview.astro
```diff
@@ -1,24 +1,32 @@
...
 
 const section = getPathogenMegaMenuSections()[organism];
+const lengthFactor = lengthScaleFactor ?? 1;
```
JonasKellerer, line 20: We can then remove this line here.

website/src/components/views/overview/GenericOverview.astro
```diff
@@ -42,5 +50,34 @@ const breadcrumbs = [...defaultBreadcrumbs, ...paths[organism].breadcrumbs];
...
+    <div class='mt-8'>
+    {addGenomeDataViewer && <ComponentWrapper title='Genome Data Viewer'>
+        <div class='flex flex-wrap gap-4'>
+            {
+                genome.map((gene) => (
+                    <div
+                        class='flex flex-col'
+                        style={{
+                            width: (gene.genomeLength / maxLength / lengthFactor) * 100 + '%',
+                            height: '100%',
```
JonasKellerer, line 62: We can add this to the class above `h-full`.

website/src/components/views/overview/GenericOverview.astro
```diff
@@ -1,24 +1,32 @@
...
 import ContaineredPageLayout from '../../../layouts/ContaineredPage/ContaineredPageLayout.astro';
+import { getGenomeDataSections, type GenomeData } from '../../../layouts/base/header/getGenomeData';
 import { getPathogenMegaMenuSections } from '../../../layouts/base/header/getPathogenMegaMenuSections';
 import { PageHeadline } from '../../../styles/containers/PageHeadline';
 import { paths } from '../../../types/Organism';
+import ComponentWrapper from '../../ComponentWrapper.astro';
 import { iconMapping } from '../../iconCss';
 
 type Props = {
     organism: keyof ReturnType<typeof getPathogenMegaMenuSections>;
```
JonasKellerer, line 12: We could than also use the returnType of `getGenomeDataSections` instead.

website/src/types/Organism.ts
```diff
@@ -73,6 +231,7 @@ export const organismConfig = {
         backgroundColorFocus: 'group-hover:bg-indigo',
         menuListEntryDecoration: 'decoration-indigo',
         borderEntryDecoration: 'hover:border-indigo',
+        genome: [],
```
JonasKellerer, line 234: We could remove this line and take care of this case in the `getGenomeDataSections` function. If it is not defined, the enrty will not show in the returned object.

website/src/types/Organism.ts
```diff
@@ -37,6 +44,7 @@ export const organismConfig = {
         backgroundColorFocus: 'group-hover:bg-cyan',
         menuListEntryDecoration: 'decoration-cyan',
         borderEntryDecoration: 'hover:border-cyan',
+        genome: [],
```
JonasKellerer, line 47: Same as below.

JonasKellerer (changes requested):
Nice. Looks great. 

I think you missed the headline for influenca A and B. 
grafik

However, I think we could make this a bit better, by putting the genome viewer display in its own astro component and add a slot to the generic overview. My comments are mostly targeted to that change, so no worries.

The component works great by the way.

website/src/types/Organism.ts
anna-parker: Thanks!

website/src/components/views/overview/GenericOverview.astro
anna-parker: I hope this is better I made all the suggestions but still kept `getGenomeDataSections` :-)

website/src/types/Organism.ts
```diff
@@ -28,6 +28,13 @@ export const organismConfig = {
         backgroundColor: 'bg-sandMuted',
         backgroundColorFocus: 'group-hover:bg-sand',
         borderEntryDecoration: 'hover:border-sand',
+        genome: [
+            {
+                gff3Source:
+                    'https://raw.githubusercontent.com/nextstrain/nextclade_data/refs/heads/master/data/nextstrain/sars-cov-2/wuhan-hu-1/orfs/genome_annotation.gff3',
```
fengelniederhammer, line 34: Did you discuss whether it makes sense to include those files into this repo and download them from the Astro server rather than from GitHub?

Do we expect frequent changes in these files? If yes, we should keep the link to GitHub. Otherwise I think it would make sense that we host them ourselves (similar to as we do for the map data already).

website/src/components/views/genomeView/GenomeView.astro
```diff
@@ -0,0 +1,44 @@
...
+
+const { organism, lengthScaleFactor = 1 } = Astro.props;
+
+const genome: GenomeData[] = getGenomeDataSections()[organism];
+const maxLength = Math.max(...genome.map((gene) => gene.genomeLength));
+
+const currentUrl = Astro.url;
+---
+
+<ComponentWrapper title='Genome Data Viewer'>
```
JonasKellerer, line 19: I would wrap this in this div, so we make the design a bit less compact. `<div class='mt-4'>`

**Before:**
grafik


**After:**
grafik

website/src/components/views/genomeView/GenomeView.astro
JonasKellerer, line 1: Nice. Thank you.

JonasKellerer (changes requested):
Thank you for the refactoring. Looks great now. 😄 
Only two minor things are left from my side: moving some files out of the public folder, and we need to update the components version, so we have the viewer.

JonasKellerer (approved):
Nice. Thank you. Looks great. 

Just a reminder: don't forget to squash the commits :)

PR #634, opened 2025-03-18
refactor(website): move values from the config to the organism constants

### LINKED ISSUE #593: Refactor the `config`
Currently we name every field that we can filter by in the config but now that we offer more filter options it would be good to extend the config to make this more generic 
This is how our config currently looks like: https://github.com/GenSpectrum/dashboards/blob/main/backend/src/main/resources/application-dashboards-prod.yaml

I would propose moving to a more generic option, similar to the Loculus config:
```
dashboards:
  organisms:
    covid:
      lapis:
        url: "https://lapis.cov-spectrum.org/open/v2"
        fields:
          - lapisField: "hostNameScientific"
            label: "host"
            type: text
            initialSearchOption: true
          - lapisField: "originatingLab"
            label: "originating lab"
            type: text
            initialSearchOption: false
          - lapisField: "region"
            label: "Region"
            type: location
            initialSearchOption: true
          - lapisField: "country"
            label: "Country"
            type: location
            initialSearchOption: true
          - lapisField: "date"
            label: "Collection Date"
            type: date
            initialSearchOption: true
```

fengelniederhammer:
resolves #593

### Summary

We only need values in the config that both website and backend need to access.

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
@@ -19,11 +19,6 @@ data class OrganismConfig(
 data class LapisConfig(
     val url: String,
     val mainDateField: String,
-    val locationFields: List<String>,
-    val authorsField: String?,
-    val authorAffiliationsField: String?,
-    val originatingLabField: String?,
-    val submittingLabField: String?,
     val additionalFilters: Map<String, String>?,
 )
 
--- a/backend/src/main/resources/application-dashboards-prod.yaml
+++ b/backend/src/main/resources/application-dashboards-prod.yaml
@@ -4,14 +4,6 @@ dashboards:
       lapis:
         url: "https://lapis.cov-spectrum.org/open/v2"
         mainDateField: "date"
-        locationFields:
-          - "region"
-          - "country"
-          - "division"
-        originatingLabField: "originatingLab"
-        submittingLabField: "submittingLab"
-        accessionDownloadFields:
-          - "strain"
       externalNavigationLinks:
         - url: "https://cov-spectrum.org"
           label: "CoV-Spectrum"
@@ -20,23 +12,9 @@ dashboards:
       lapis:
         url: "https://api.loculus.genspectrum.org/h5n1"
         mainDateField: "sampleCollectionDate"
-        locationFields:
-          - "country"
-          - "division"
-        authorAffiliationsField: "authorAffiliations"
-        authorsField: "authors"
         additionalFilters:
           versionStatus: "LATEST_VERSION"
           isRevocation: "false"
-        accessionDownloadFields:
-          - "insdcAccessionFull_seg1"
-          - "insdcAccessionFull_seg2"
-          - "insdcAccessionFull_seg3"
-          - "insdcAccessionFull_seg4"
-          - "insdcAccessionFull_seg5"
-          - "insdcAccessionFull_seg6"
-          - "insdcAccessionFull_seg7"
-          - "insdcAccessionFull_seg8"
       externalNavigationLinks:
         - url: "https://loculus.genspectrum.org/h5n1"
           label: "Browse data"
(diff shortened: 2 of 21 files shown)


backend/src/main/resources/application-dashboards-prod.yaml
```diff
@@ -4,14 +4,6 @@ dashboards:
...
         mainDateField: "date"
-        locationFields:
-          - "region"
-          - "country"
-          - "division"
-        originatingLabField: "originatingLab"
-        submittingLabField: "submittingLab"
-        accessionDownloadFields:
-          - "strain"
       externalNavigationLinks:
```
anna-parker, line 7: it looks like you still have `externalNavigationLinks` in the config?

anna-parker (changes requested):
I think there might still be an issue with the config

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
```diff
@@ -13,22 +13,10 @@ data class DashboardsConfig(
 
 data class OrganismConfig(
     val lapis: LapisConfig,
-    val externalNavigationLinks: List<ExternalNavigationLink>?,
```
JonasKellerer, line 16: I think we still need them here. Or at least, I get an error from my idea:

grafik

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
JonasKellerer, line 1: Maybe not this PR, but we could also update the Organism in the backend. Currently, "h1n1pdm", "h3n2", "influenzaA", "influenzaB", "victoria", "ebolaSudan", "ebolaZaire" and "cchf" are missing.

website/src/views/ebolaSudan.ts
```diff
@@ -49,13 +54,13 @@ class EbolaSudanConstants implements OrganismConstants {
...
-    public readonly locationFields: string[];
+    public readonly locationFields = PATHOPLEXUS_LOCATION_FIELDS;
     public readonly lineageFilters: LineageFilterConfig[] = [];
     public readonly useAdvancedQuery = false;
     public readonly hostField: string = hostField;
-    public readonly authorsField: string | undefined;
-    public readonly authorAffiliationsField: string | undefined;
-    public readonly accessionDownloadFields;
+    public readonly authorsField = 'authors';
+    public readonly authorAffiliationsField = 'authorAffiliations';
```
JonasKellerer, line 62: I think we use both fields in all pathoplexus instance. So I would be in favor of putting them also into a constant, like "PATHOPLEXUS_LOCATION_FIELDS"

JonasKellerer (approved):
Nice to have the config more in one place. 

I'm not sure if this is a good idea, but we could split the config into two. One that is used for both backend and frontend and one that is used for the frontend (for the external navigation links). But it seems this is too much effort, for such a small thing. Maybe we keep this in mind, when we have more instance specific configs only for the frontend.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
fengelniederhammer: IMO that's too much effort for now. It would require adapting backend tests, too.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
fengelniederhammer: I don't quite agree that we "need" it, since it's only the IDE that complains that it doesn't know what to do with those properties. I added it nonetheless.

website/vitest.setup.ts
```diff
@@ -165,374 +165,136 @@ export const testOrganismsConfig = {
...
-                'insdcAccessionFull_seg2',
-                'insdcAccessionFull_seg3',
-                'insdcAccessionFull_seg4',
-                'insdcAccessionFull_seg5',
-                'insdcAccessionFull_seg6',
-                'insdcAccessionFull_seg7',
-                'insdcAccessionFull_seg8',
-            ],
         },
-        externalNavigationLinks: [
```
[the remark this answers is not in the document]
fengelniederhammer: They should not be necessary for unit tests, that's why I removed them.

PR #640, opened 2025-03-21
feat(website): add NAI mutation annotations

anna-parker:
resolves #

### DIFF
--- a/website/src/pages/swiss-wastewater/flu.astro
+++ b/website/src/pages/swiss-wastewater/flu.astro
@@ -6,7 +6,12 @@ import GsWastewaterMutationsOverTime from '../../components/genspectrum/GsWastew
 import { defaultBreadcrumbs } from '../../layouts/Breadcrumbs';
 import DataPageLayout from '../../layouts/OrganismPage/DataPageLayout.astro';
 import { dataOrigins } from '../../types/dataOrigins';
-import { wastewaterConfig, wastewaterPathFragment } from '../../types/wastewaterConfig';
+import {
+    mutationAnnotationsN1,
+    mutationAnnotationsN2,
+    wastewaterConfig,
+    wastewaterPathFragment,
+} from '../../types/wastewaterConfig';
 
 const lapisUrl = `${wastewaterConfig.lapisBaseUrl}/influenza`;
 const getReferencesUrl = `${lapisUrl}/sample/aggregated?fields=reference&versionStatus=LATEST_VERSION&isRevocation=false`;
@@ -28,32 +33,35 @@ const references = (await (await fetch(getReferencesUrl)).json()).data.map(
     dataOrigins={[dataOrigins.wise]}
     lapisUrl={lapisUrl}
 >
-    <gs-app lapis={lapisUrl}>
-        <div class='m-4 flex flex-col gap-4'>
-            {
-                references.map((reference) => (
-                    <>
-                        <h2 class='text-2xl'>{reference}</h2>
-                        {(['nucleotide', 'amino acid'] satisfies SequenceType[]).map((sequenceType) => (
-                            <GsWastewaterMutationsOverTime
-                                title={`${sequenceType} mutations`}
-                                lapisFilter={{
-                                    reference,
-                                    versionStatus: 'LATEST_VERSION',
-                                    isRevocation: false,
-                                }}
-                                sequenceType={sequenceType}
-                            >
-                                <InfluenzaWastewaterInfo
-                                    slot='infoText'
-                                    reference={reference}
+    {
+        references.map((reference) => {
+            const mutationAnnotations = reference === 'N2' ? mutationAnnotationsN2 : mutationAnnotationsN1;
+            return (
+                <gs-app lapis={lapisUrl} mutationAnnotations={JSON.stringify(mutationAnnotations)}>
+                    <div class='m-4 flex flex-col gap-4'>
+                        <>
+                            <h2 class='text-2xl'>{reference}</h2>
+                            {(['nucleotide', 'amino acid'] satisfies SequenceType[]).map((sequenceType) => (
+                                <GsWastewaterMutationsOverTime
+                                    title={`${sequenceType} mutations`}
+                                    lapisFilter={{
+                                        reference,
+                                        versionStatus: 'LATEST_VERSION',
+                                        isRevocation: false,
+                                    }}
                                     sequenceType={sequenceType}
-                                />
-                            </GsWastewaterMutationsOverTime>
-                        ))}
-                    </>
-                ))
-            }
-        </div>
-    </gs-app>
+                                >
+                                    <InfluenzaWastewaterInfo
+                                        slot='infoText'
+                                        reference={reference}
+                                        sequenceType={sequenceType}
+                                    />
+                                </GsWastewaterMutationsOverTime>
+                            ))}
+                        </>
+                    </div>
+                </gs-app>
+            );
+        })
+    }
 </DataPageLayout>
--- a/website/src/types/wastewaterConfig.ts
+++ b/website/src/types/wastewaterConfig.ts
@@ -1,3 +1,5 @@
+import type { MutationAnnotation } from '@genspectrum/dashboard-components/util';
+
 export const wastewaterPathFragment = 'swiss-wastewater';
 
 export const wastewaterConfig = {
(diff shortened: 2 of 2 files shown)


website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,193 @@ export const wastewaterConfig = {
...
+            'I117R',
+        ],
+    },
+];
+
+// WHO uses reference KJ609208.1 and AB124658.1, mutations here use reference CY114383.1
+// When KJ609208.1 and AB124658.1 are aligned to CY114383.1 using nextclade we see no insertions/deletions only mutations
+// Therefore, we can use CY114383.1 as the reference for the mutations
+// See https://github.com/anna-parker/NAIMutations/tree/main for calculation details
+// TODO: Some mutations are only of interest when with other mutations (e.g. deletions) but are now marked individually
```
fengelniederhammer, line 136: Is that relevant for users? If yes, we could put it on the `swiss-wastewater/index.astro` page, because we wanted to use that for docs.

website/src/pages/swiss-wastewater/flu.astro
```diff
@@ -28,32 +33,35 @@ const references = (await (await fetch(getReferencesUrl)).json()).data.map(
...
-                                    isRevocation: false,
-                                }}
-                                sequenceType={sequenceType}
-                            >
-                                <InfluenzaWastewaterInfo
-                                    slot='infoText'
-                                    reference={reference}
+    {
+        references.map((reference) => {
+            const mutationAnnotations = reference === 'N2' ? mutationAnnotationsN2 : mutationAnnotationsN1;
```
fengelniederhammer, line 38: Is that logic here correct? If the reference is not "N2", then use the annotations of "N1"? Especially since we don't have control over which `references` we have (because we fetch that from LAPIS)?

Wouldn't it be safer to put that logic into a function in `wastewaterConfig.ts` that does a 
```typescript
switch (reference) {
  case ...:
    return ...;
  default:
    return [];
}
```

anna-parker: ah yes this is a better approach - I will change, thansk for spotting!

website/src/types/wastewaterConfig.ts
anna-parker: I'm not sure if users need to know - I would say probably not. I think its good to keep the description here in case we need to make changes in future as the calculation isn't trivial - but I think it is fine to just keep it as a comment for now :-)

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+            return mutationAnnotationsN1;
+        case 'N2':
+            return mutationAnnotationsN2;
+        default:
+            return [];
+    }
+}
+
+// See https://github.com/anna-parker/NAIMutations/tree/main for calculation details
+// TODO: Some mutations are only of interest when with other mutations (e.g. deletions) but are now marked individually
```
JonasKellerer, line 34: Is there a ticket that we can link here? It would be nice to keep track of this.

JonasKellerer: Is there a way to have an annotation for each of the "mutations that are only of interest  when with other mutations", or are those to many? I could imagine something like: 
"description: This mutation is asociated with ... and is only relevant when paired with mutations A123T"

website/src/pages/swiss-wastewater/flu.astro
```diff
@@ -6,7 +6,7 @@ import GsWastewaterMutationsOverTime from '../../components/genspectrum/GsWastew
 import { defaultBreadcrumbs } from '../../layouts/Breadcrumbs';
 import DataPageLayout from '../../layouts/OrganismPage/DataPageLayout.astro';
 import { dataOrigins } from '../../types/dataOrigins';
-import { wastewaterConfig, wastewaterPathFragment } from '../../types/wastewaterConfig';
+import { getMutationAnnotation, wastewaterConfig, wastewaterPathFragment } from '../../types/wastewaterConfig';
 
 const lapisUrl = `${wastewaterConfig.lapisBaseUrl}/influenza`;
 const getReferencesUrl = `${lapisUrl}/sample/aggregated?fields=reference&versionStatus=LATEST_VERSION&isRevocation=false`;
```
JonasKellerer, line 12: Can't we do the same we do for rsv? There we provide a list of references and use them. Is there a benefit in calling lapis for the references? The benefit would be, that we can type it and use that in `getMutationAnnotation`.

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+            'I117R',
+        ],
+    },
+];
+
+// WHO uses reference KJ609208.1 and AB124658.1, mutations here use reference CY114383.1
+// When KJ609208.1 and AB124658.1 are aligned to CY114383.1 using nextclade we see no insertions/deletions only mutations
+// Therefore, we can use CY114383.1 as the reference for the mutations
+// See https://github.com/anna-parker/NAIMutations/tree/main for calculation details
+// TODO: Some mutations are only of interest when with other mutations (e.g. deletions) but are now marked individually
```
JonasKellerer, line 147: Same comment as above.

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+            '248-',
+            '245-',
+            '250-',
+        ],
+    },
+    {
+        name: 'Peramivir resistance mutations',
+        description:
+            "This mutation is associated with reduced inhibition by Peramivir, for more details see the <a class='link' href='https://www.who.int/teams/global-influenza-programme/laboratory-network/quality-assurance/antiviral-susceptibility-influenza/neuraminidase-inhibitor'>Global Influenza Programme Report</a>.",
+        symbol: '*',
```
JonasKellerer, line 192: This is the same symbol as 'Laninamivir resistance mutations'. Maybe we could use another one?

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+        nucleotideMutations: [],
+        aminoAcidMutations: ['N142S', 'E119V', 'R292K', 'D151A', 'T148I'],
+    },
+    {
+        name: 'Zanamivir resistance mutations',
+        description:
+            "This mutation is associated with reduced inhibition by Zanamivir, for more details see the <a class='link' href='https://www.who.int/teams/global-influenza-programme/laboratory-network/quality-assurance/antiviral-susceptibility-influenza/neuraminidase-inhibitor'>Global Influenza Programme Report</a>.",
+        symbol: '*',
```
JonasKellerer, line 200: This is the same symbol as 'Laninamivir resistance mutations'. Maybe we could use another one?

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+            "This mutation is associated with reduced inhibition by Laninamivir, for more details see the <a class='link' href='https://www.who.int/teams/global-influenza-programme/laboratory-network/quality-assurance/antiviral-susceptibility-influenza/neuraminidase-inhibitor'>Global Influenza Programme Report</a>.",
+        symbol: '*',
+        nucleotideMutations: [],
+        aminoAcidMutations: ['T148I', 'E119V', 'N142S'],
+    },
+    {
+        name: 'Oseltamivir resistance mutations',
+        description:
+            "This mutation is associated with reduced inhibition by Oseltamivir, for more details see the <a class='link' href='https://www.who.int/teams/global-influenza-programme/laboratory-network/quality-assurance/antiviral-susceptibility-influenza/neuraminidase-inhibitor'>Global Influenza Programme Report</a>.",
+        symbol: '*',
```
JonasKellerer, line 161: This is the same symbol as 'Laninamivir resistance mutations'. Maybe we could use another one?

website/src/types/wastewaterConfig.ts
JonasKellerer: Do we want to use the same symbols we also use for N1? So my question is, do we want to "train" our users to spotting these specific annotations or are these annotations inheritly different?

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,204 @@ export const wastewaterConfig = {
...
+    }
+}
+
+// See https://github.com/anna-parker/NAIMutations/tree/main for calculation details
+// TODO: Some mutations are only of interest when with other mutations (e.g. deletions) but are now marked individually
+const mutationAnnotationsN1: MutationAnnotation[] = [
+    {
+        name: 'Laninamivir resistance mutations',
+        description:
+            "This mutation is associated with reduced inhibition by Laninamivir, for more details see the <a class='link' href='https://www.who.int/teams/global-influenza-programme/laboratory-network/quality-assurance/antiviral-susceptibility-influenza/neuraminidase-inhibitor'>Global Influenza Programme Report</a>.",
```
JonasKellerer, line 39: It feels that we have this description multiple times below, only the "name" exchanged. It would be nice to have this in a function with the name as a parameter. This would make it easier to fix the text later on, when we want to change it.

JonasKellerer (changes requested):
Nice. Good to have these annotations. Only some smaller things.

website/src/types/wastewaterConfig.ts
anna-parker: good catch - I forgot to make them unique for medication

website/src/pages/swiss-wastewater/flu.astro
anna-parker: I think the idea was we could get more influenza subtypes but there will most likely never be anymore rsv subtypes (unless the scientific community creates a new naming system). But I can change this.

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,177 @@ export const wastewaterConfig = {
 export const RSVTypes = ['RSV-A', 'RSV-B'] as const;
 
 export type RSVType = (typeof RSVTypes)[number];
+
+export const InfluenzaTypes = ['H1', 'N1', 'H3', 'N2'] as const;
+
+export type InfluenzaType = (typeof InfluenzaTypes)[number];
+
+export function getMutationAnnotation(reference: string): MutationAnnotation[] {
```
JonasKellerer, line 26:
```suggestion
export function getMutationAnnotation(reference: InfluencaType): MutationAnnotation[] {
```

website/src/types/wastewaterConfig.ts
```diff
@@ -16,3 +18,177 @@ export const wastewaterConfig = {
...
+        case 'N1':
+            return mutationAnnotationsN1;
+        case 'N2':
+            return mutationAnnotationsN2;
+        default:
+            return [];
+    }
+}
+
+function symbolMap(medication: string): string {
```
JonasKellerer, line 37: Nitpick: We could also type the medications to fixed values.

JonasKellerer (approved):
Nice. Thank you!

Only squash the commits to one, but I think this will be done by "Squash and Merge" right?

Issue #618, opened 2025-03-11
Put location filter config into the page state selectors


fengelniederhammer:
You have now moved the config of the baseline and variant filter inside the component and read it from the view. We could also just provide the view and use this for all other things. [...]

_Originally posted by @JonasKellerer in https://github.com/GenSpectrum/dashboards/pull/606#discussion_r1983091360_
            
As a follow up to #606, move the `locationFilterConfig` into the page state selectors.

PR #663, opened 2025-04-08
feat(components): allow for multiple location filter

JonasKellerer:
Resolves: #618

### Summary

Moves the location config to the other configs (text and date). This allows us to set more than one location filter. 

### Screenshot

grafik


fengelniederhammer (approved):
LGTM

PR #666, opened 2025-04-10
feat(components): use influenza a and b as breadcrumbs

### LINKED ISSUE #653: Add influenza overview index page to breadcrumb nav
On a page of an influenza subtype like https://genspectrum.org/influenza-a/h5n1 and https://genspectrum.org/influenza-a/h5n1/sequencing-efforts, the breadcrumb navigation starts with

```
Home > Influenza A/H5N1 
```

I think that it would be nice to turn it to the following instead:

```
Home > Influenza A > H5N1
```

With `Influenza A` linking to the index page of [influenza A](https://genspectrum.org/influenza-a). The same applies to influenza B.

JonasKellerer:
Resolves #653 
### Summary

- uses new paths object to define paths to the pages
- I put it not in `OrganismConstants`, since we need the breadcrumbs also on the index pages, where they are not known

### Screenshot

grafik

### DIFF
--- a/website/src/types/Organism.ts
+++ b/website/src/types/Organism.ts
@@ -1,5 +1,7 @@
 import { z } from 'zod';
 
+import { wastewaterBreadcrumb, wastewaterPathFragment } from './wastewaterConfig.ts';
+
 export const Organisms = {
     covid: 'covid' as const,
     influenzaA: 'influenzaA' as const,
@@ -38,8 +40,8 @@ export const organismConfig = {
     },
     [Organisms.h5n1]: {
         organism: Organisms.h5n1,
-        pathFragment: 'influenza-a/h5n1',
-        label: 'Influenza A/H5N1',
+        pathFragment: 'h5n1',
+        label: 'H5N1',
         backgroundColor: 'bg-cyanMuted',
         backgroundColorFocus: 'group-hover:bg-cyan',
         menuListEntryDecoration: 'decoration-cyan',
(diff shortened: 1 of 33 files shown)


website/src/types/Organism.ts
```diff
@@ -145,6 +145,161 @@ export const organismConfig = {
...
+                name: organismConfig[Organisms.cchf].label,
+                href: `/${organismConfig[Organisms.cchf].pathFragment}`,
+            },
+        ],
+    },
+    swissWastewater: {
+        basePath: `/swiss-wastewater`,
+        breadcrumbs: [
+            {
+                name: organismConfig[Organisms.cchf].label,
```
fengelniederhammer, line 296: Copy-paste error ;)

website/src/views/h3n2.ts
```diff
@@ -127,11 +127,7 @@ export class H3n2CompareSideBySideView extends BaseView<
 
         super(
             constants,
-            new CompareSideBySideStateHandler(
-                constants,
-                defaultPageState,
-                organismConfig[constants.organism].pathFragment,
```
fengelniederhammer, line 133: Some organisms are missing this change - I'll adapt it such that the state handlers simply get the values or their own.

website/src/types/Organism.ts
```diff
@@ -145,6 +145,161 @@ export const organismConfig = {
...
+            {
+                name: organismConfig[Organisms.influenzaA].label,
+                href: `/${organismConfig[Organisms.influenzaA].pathFragment}`,
+            },
+            {
+                name: organismConfig[Organisms.h5n1].label,
+                href: `/${organismConfig[Organisms.influenzaA].pathFragment}/${organismConfig[Organisms.h5n1].pathFragment}`,
+            },
+        ],
+    },
```
fengelniederhammer, line 162: I think we can extract some common parts into functions.

fengelniederhammer (approved):

PR #667, opened 2025-04-13
chore(website): update @genspectrum/dashboard-components to 0.19.0

fengelniederhammer:


fengelniederhammer:
I created https://github.com/capricorn86/happy-dom/issues/1802. Let's see whether there is a better fix than switching to happy-dom.

chaoran-chen (approved):

PR #686, opened 2025-05-05
feat(website): add completeness filter

### LINKED ISSUE #637: Add completeness filters
Once we have  https://github.com/GenSpectrum/dashboard-components/issues/349)
Leftover from #550

Add completeness filter to Pathoplexus instances and other instances, that are not COVID.

JonasKellerer:
Resolves #637

### Summary

- uses the gs-number-range-filter
- adds new filter type "number"
- adds the new filter to multiple organisms (west-nile, rsvA, rsvB, mpox, ebola sudan, ebola zaire), which have one field named "completeness". Most other organisms have more than one field, eg "completeness_seg1", etc. 

### Screenshot
grafik

### DIFF
--- a/website/src/components/genspectrum/GsNumerRangeFilter.tsx
+++ b/website/src/components/genspectrum/GsNumerRangeFilter.tsx
@@ -1,0 +1,52 @@
+import type { NumberRange } from '@genspectrum/dashboard-components/util';
+import { useEffect, useRef } from 'react';
+import '@genspectrum/dashboard-components/components';
+
+export function GsNumerRangeFilter({
+    lapisField,
+    width,
+    onNumberRangeChanged = () => {},
+    value,
+    sliderMin,
+    sliderMax,
+    sliderStep,
+}: {
+    lapisField: string;
+    width?: string;
+    onNumberRangeChanged?: (numberRange: NumberRange) => void;
+    value?: NumberRange;
+    sliderMin?: number;
+    sliderMax?: number;
+    sliderStep?: number;
+}) {
+    const numberRangeFilterRef = useRef<HTMLElement>();
+
+    useEffect(() => {
+        const handleNumberRangeFilterChange = (event: CustomEvent) => {
+            onNumberRangeChanged(event.detail);
+        };
+
+        const currentInputRef = numberRangeFilterRef.current;
+        if (currentInputRef !== undefined) {
+            currentInputRef.addEventListener('gs-number-range-value-changed', handleNumberRangeFilterChange);
+        }
+
+        return () => {
+            if (currentInputRef !== undefined) {
+                currentInputRef.removeEventListener('gs-number-range-value-changed', handleNumberRangeFilterChange);
+            }
+        };
+    }, [onNumberRangeChanged]);
+
+    return (
+        <gs-number-range-filter
+            ref={numberRangeFilterRef}
+            lapisField={lapisField}
+            width={width}
+            value={JSON.stringify(value ?? {})}
+            sliderMin={sliderMin}
+            sliderMax={sliderMax}
+            sliderStep={sliderStep}
+        ></gs-number-range-filter>
+    );
+}
--- a/website/src/views/OrganismConstants.ts
+++ b/website/src/views/OrganismConstants.ts
@@ -251,6 +251,15 @@ export function getPathoplexusFilters({
     ];
 }
 
+export const COMPLETENESS_FILTER = {
+    lapisField: 'completeness',
+    type: 'number' as const,
+    label: 'Completeness',
+    sliderMin: 0,
+    sliderMax: 1,
+    sliderStep: 0.01,
+};
+
 export const GENPSECTRUM_LOCULUS_HOST_FIELD = 'hostNameScientific';
 
 export function getGenspectrumLoculusFilters({
--- a/website/src/views/helpers.ts
+++ b/website/src/views/helpers.ts
@@ -1,4 +1,4 @@
-import type { DateRangeOption } from '@genspectrum/dashboard-components/util';
+import type { DateRangeOption, NumberRange } from '@genspectrum/dashboard-components/util';
 
 import type { VariantFilter } from './View.ts';
 import type { MutationFilter } from '../components/genspectrum/GsMutationFilter.tsx';
@@ -41,6 +41,17 @@ export const setSearchFromDateRange = (
         search.set(name, serializedValue);
     }
 };
+
+export function setSearchFromNumberRange(search: URLSearchParams, name: string, numberRange: NumberRange | undefined) {
+    if (numberRange !== undefined) {
+        if (numberRange.min) {
+            search.set(`${name}From`, String(numberRange.min));
+        }
+        if (numberRange.max) {
+            search.set(`${name}To`, String(numberRange.max));
+        }
+    }
+}
 
 export const getStringFromSearch = (
     search: URLSearchParams | Map<string, string>,
@@ -85,6 +96,23 @@ export const getDateRangeFromSearch = (
     }
     return undefined;
 };
+
+export function getNumberRangeFromSearch(
+    search: URLSearchParams | Map<string, string>,
+    name: string,
+): NumberRange | undefined {
+    const minString = search.get(`${name}From`);
+    const maxString = search.get(`${name}To`);
+
+    const min = minString ? Number(minString) : undefined;
+    const max = maxString ? Number(maxString) : undefined;
+
+    if (min || max) {
+        return { min, max };
+    }
+
+    return undefined;
+}
 
 export type LapisLocation = Record<string, string | undefined>;
 
--- a/website/src/views/influenza-a.ts
+++ b/website/src/views/influenza-a.ts
@@ -86,6 +86,7 @@ const defaultDatasetFilter: DatasetFilter = {
     dateFilters: {
         [GENSPECTRUM_LOCULUS_MAIN_FILTER_DATE_COLUMN]: dateRangeOptionPresets.lastYear,
     },
+    numberFilters: {},
 };
 
 export class InfluenzaACompareSideBySideView extends BaseView<
--- a/website/src/views/pageStateHandlers/PageStateHandler.ts
+++ b/website/src/views/pageStateHandlers/PageStateHandler.ts
@@ -1,13 +1,15 @@
-import type { DateRangeOption, LapisFilter } from '@genspectrum/dashboard-components/util';
+import type { DateRangeOption, LapisFilter, NumberRange } from '@genspectrum/dashboard-components/util';
 
 import type { BaselineFilterConfig } from '../../components/pageStateSelectors/BaselineSelector.tsx';
 import { type Dataset, type Id, type VariantFilter } from '../View.ts';
 import {
     getDateRangeFromSearch,
     getLapisLocationFromSearch,
+    getNumberRangeFromSearch,
     getStringFromSearch,
     type LapisLocation,
     setSearchFromDateRange,
+    setSearchFromNumberRange,
     setSearchFromString,
 } from '../helpers.ts';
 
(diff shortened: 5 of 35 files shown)


website/src/views/rsvA.ts
```diff
@@ -56,6 +57,7 @@ class RsvAConstants implements OrganismConstants {
             dateRangeOptions: fineGrainedDefaultDateRangeOptions,
             earliestDate,
         }),
+        COMPLETENESS_FILTER,
```
fengelniederhammer, line 60: I think we could instead add this to `getGenspectrumLoculusFilters` (and `getPathoplexusFilters`). If no additional info is passed, then it could simply return the single-segmented completeness. Otherwise we could pass some basic config like
```
[ { segmentName: 'seg1', label: 'Segment 1' } , { ... } , ... ]
```
to those functions when they need to generate completeness filters for multiple segments.

website/src/components/genspectrum/GsNumerRangeFilter.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const numberRangeFilterRef = useRef<HTMLElement>();
+
+    useEffect(() => {
+        const handleNumberRangeFilterChange = (event: CustomEvent) => {
+            onNumberRangeChanged(event.detail);
+        };
+
+        const currentInputRef = numberRangeFilterRef.current;
+        if (currentInputRef !== undefined) {
+            currentInputRef.addEventListener('gs-number-range-value-changed', handleNumberRangeFilterChange);
```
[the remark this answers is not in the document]
JonasKellerer: Could you elaborate on the benefit? This would then just be an empty event listener?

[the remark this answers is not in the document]
JonasKellerer: True. We don't use them for the other components. I'll create a separate PR for that.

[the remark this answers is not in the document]
JonasKellerer: We don't expose these functions for the other components. I'll add this in a separate PR for all the components.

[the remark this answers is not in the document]
JonasKellerer: https://github.com/GenSpectrum/dashboards/pull/691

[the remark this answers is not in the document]
JonasKellerer: I wanted to add this as a separate PR, but i made a mistake. It is now added to this PR.

website/src/views/rsvA.ts
JonasKellerer: As discussed: We will display only segments 4 and 6 for influenca, and for cchf, we will dasplay all completeness filters.

website/src/views/helpers.ts
```diff
@@ -86,6 +97,23 @@ export const getDateRangeFromSearch = (
...
+    search: URLSearchParams | Map<string, string>,
+    name: string,
+): NumberRange | undefined {
+    const minString = search.get(`${name}From`);
+    const maxString = search.get(`${name}To`);
+
+    const min = minString ? Number(minString) : undefined;
+    const max = maxString ? Number(maxString) : undefined;
+
+    if ((min !== undefined && !Number.isNaN(min)) || (max !== undefined && !Number.isNaN(max))) {
```
fengelniederhammer, line 110: I just discovered this function. Does it work? According to the docs, it should do the same.
```suggestion
    if (Number.isFinite(min) || Number.isFinite(max)) {
```

website/src/views/pageStateHandlers/PageStateHandler.ts
```diff
@@ -66,10 +68,37 @@ export function toLapisFilterWithoutVariant(
...
+                ? {
+                      [`${lapisField}To`]: String(numberRange.max),
+                  }
+                : {};
+
+            return {
+                ...acc,
+                ...fromValue,
+                ...toValue,
+            };
```
fengelniederhammer, line 92:
```suggestion
            return {
                ...acc,
                [`${lapisField}From`]: numberRange.min === undefined ? undefined : String(numberRange.min),
                [`${lapisField}To`]: numberRange.max === undefined ? undefined : String(numberRange.max),
            };
```

Doesn't that already do the job?

(We could even get rid of the early return by using `numberRange?.min`)

website/src/views/influenza-a.ts
```diff
@@ -53,10 +54,11 @@ class InfluenzaAConstants implements OrganismConstants {
         ...getGenspectrumLoculusFilters({
             dateRangeOptions: fineGrainedDefaultDateRangeOptions,
             earliestDate,
+            completenessSuffixes: INFLUNEZA_COMPLETENESS_SUFFIXES,
```
fengelniederhammer, line 57: I just noticed that Flu A doesn't have any completeness data:

```
POST https://api.loculus.genspectrum.org/influenza-a/sample/aggregated
Accept: */*
Content-Type: application/json

{
  "fields": [
    "completeness_seg4",
    "completeness_seg6"
  ]
}
```

returns

```
{
  "data": [
    {
      "count": 311514,
      "completeness_seg4": null,
      "completeness_seg6": null
    }
  ],
  "info": {
    "dataVersion": "1743684215",
    "requestId": "33df9e2c-b541-466b-b517-4aae011c1420",
    "requestInfo": "Influenza A on api.loculus.genspectrum.org at 2025-05-14T07:14:43.011213516",
    "reportTo": "Please report to https://github.com/GenSpectrum/LAPIS/issues in case you encounter any unexpected issues. Please include the request ID and the requestInfo in your report.",
    "lapisVersion": "0.4.1",
    "siloVersion": "0.5.7"
  }
}
```

Thus setting any completeness filter will yield no data:
image

@anna-parker is there something wrong with the data/preprocessing? Is that expected? 
With the current data, I think it's better to have no completeness filters for Flu A.

website/src/views/pageStateHandlers/PageStateHandler.ts
```diff
@@ -66,10 +68,37 @@ export function toLapisFilterWithoutVariant(
         {} as { [key: string]: string | undefined },
     );
 
+    const numberRangeFilters = Object.entries(pageState.datasetFilter.numberFilters).reduce(
+        (acc, [lapisField, numberRange]) => {
+            if (numberRange === undefined) {
+                return acc;
+            }
+
+            const fromValue = numberRange.min
```
fengelniederhammer, line 77: Anyway, this should check for `!== undefined`, because `0` is also falsy. (That's why I think it's better to always be explicit and never just rely on JS' truthiness inference.)

fengelniederhammer: Let's add a unit test somewhere.

website/src/views/influenza-a.ts
fengelniederhammer: And same for Flu B

fengelniederhammer (changes requested):

website/src/components/genspectrum/GsNumerRangeFilter.tsx
```diff
@@ -0,0 +1,60 @@
...
+        const currentInputRef = numberRangeFilterRef.current;
+        if (currentInputRef === undefined) {
+            return;
+        }
+
+        const handleNumberRangeFilterChange = (event: CustomEvent) => {
+            onNumberRangeChanged(event.detail);
+        };
+
+        const handleLapisFilterChange = (event: CustomEvent) => {
```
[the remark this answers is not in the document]
JonasKellerer: I approved and released it.

website/src/views/OrganismConstants.ts
```diff
@@ -248,18 +252,48 @@ export function getPathoplexusFilters({
...
+export const COMPLETENESS_FILTER = {
+    lapisField: 'completeness',
+    type: 'number' as const,
+    label: 'Completeness',
+    sliderMin: 0,
+    sliderMax: 1,
+    sliderStep: 0.01,
+};
+
+export const INFLUNEZA_COMPLETENESS_SUFFIXES = [
```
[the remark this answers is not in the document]
JonasKellerer: Good catch. Somehow my idea only shows them 50% of the time. 🙈

website/src/views/OrganismConstants.ts
```diff
@@ -248,18 +252,48 @@ export function getPathoplexusFilters({
...
         },
+        ...getCompletenessFilters(completenessSuffixes),
     ];
 }
 
-export const GENPSECTRUM_LOCULUS_HOST_FIELD = 'hostNameScientific';
+type SuffixConfig = { suffix: string; suffixLabel: string };
+
+function getCompletenessFilters(completenessSuffixConfig: SuffixConfig[]) {
+    return completenessSuffixConfig.length === 0
```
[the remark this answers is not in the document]
JonasKellerer: Good idea.

website/src/views/helpers.ts
JonasKellerer: Good idea. Looks way better now.

website/src/views/pageStateHandlers/PageStateHandler.ts
JonasKellerer: Yes and now. This way, we set a key "numberFrom" to undefined, which is then also send to lapis. I wanted to only send filters, that have a value. Therefore, I need the construct with the `{}`, which is then left out during destructuring. 

I added a test, to show this.

website/src/views/influenza-a.ts
JonasKellerer: Oh, I just looked at the available fields and not, if there are any values. I can disable the filter on those pages for now.

website/src/views/pageStateHandlers/PageStateHandler.ts
fengelniederhammer: It doesn't really matter, `undefined` is made to be "not present". When serializing an object to JSON, then all keys where the value is `undefined` will be skipped.

It would only matter if we overwrote keys. But I think it's safe to assume that in this context, we will never have duplicate keys on those objects. -> I'd prefer the shorter code which writes a couple of "useless" `undefined`s.

website/src/views/pageStateHandlers/PageStateHandler.ts
JonasKellerer: I added a unit test.

website/src/views/helpers.ts
```diff
@@ -86,6 +97,23 @@ export const getDateRangeFromSearch = (
...
+    search: URLSearchParams | Map<string, string>,
+    name: string,
+): NumberRange | undefined {
+    const minString = search.get(`${name}From`);
+    const maxString = search.get(`${name}To`);
+
+    const min = minString ? Number(minString) : undefined;
+    const max = maxString ? Number(maxString) : undefined;
+
+    if (Number.isFinite(min) || Number.isFinite(max)) {
```
fengelniederhammer, line 110: I think they need individual checks. `completeness_seg4From=0.06&completeness_seg4To=notANumber` will yield `{ min: 0.06, max: NaN }`, which is probably not what we want.

website/src/views/pageStateHandlers/dateFilterFromToUrl.ts
```diff
@@ -0,0 +1,44 @@
...
+export function parseDateRangesFromUrl(
+    search: URLSearchParams | Map<string, string>,
+    baselineFilterConfigs: BaselineFilterConfig[] | undefined,
+) {
+    const dateRangeFilterConfigs = baselineFilterConfigs?.filter((config) => config.type === 'date');
+
+    return (
+        dateRangeFilterConfigs?.reduce(
+            (acc, config) => {
+                const dateRange = getDateRangeFromSearch(search, config.dateColumn, config.dateRangeOptions);
```
fengelniederhammer, line 16: We could move `getDateRangeFromSearch` into this file, too.

fengelniederhammer: (Applies to a bunch of other functions in the helpers, too, but we can also do that later.)

fengelniederhammer (approved):
LGTM

website/src/views/helpers.ts
JonasKellerer: I added now a test, and fixed the function.

website/src/views/pageStateHandlers/PageStateHandler.ts
JonasKellerer: As discussed. We leave it like this, after some reformatting.

PR #691, opened 2025-05-13
chore(website): use event name constants instead of string

JonasKellerer:
### Summary
Follow up to #686. We now use the constant of the filter names, instead of a string. Also changes, that there is an ealry return in the useEffect inside the filters


fengelniederhammer (approved):
Thanks :)

PR #700, opened 2025-05-28
chore(website): use staging LAPIS for Covid data on staging

fengelniederhammer:
resolves #

### DIFF
--- a/backend/src/main/resources/application-dashboards-staging.yaml
+++ b/backend/src/main/resources/application-dashboards-staging.yaml
@@ -2,7 +2,7 @@ dashboards:
   organisms:
     covid:
       lapis:
-        url: "https://lapis.cov-spectrum.org/open/v2"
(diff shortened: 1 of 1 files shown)


backend/src/main/resources/application-dashboards-staging.yaml
```diff
@@ -2,7 +2,7 @@ dashboards:
   organisms:
     covid:
       lapis:
-        url: "https://lapis.cov-spectrum.org/open/v2"
+        url: "https://gs-staging-1.int.genspectrum.org/open/v2"
```
anna-parker, line 5: do we have covid data on genspectrum? I don't see it in the `organismValues.yaml`?

fengelniederhammer: We don't have it in our Loculus. The LAPIS instances run separately outside of Loculus.

PR #718, opened 2025-06-11
feat(website): add advanced query text field

### LINKED ISSUE #702: add a simple text field for advanced queries
We will likely need friendlier UI that explains the field and docs. But we can do that in a next step.

But for now let's only do the first step: Add a simple text field to the filters section. We can probably only do that for Covid for now, because we would need to update Loculus/Pathoplexus first.

JonasKellerer:
Resolves #702

### Summary

This adds the advanced query text field to each page. It is currently only enabled for covid, since we don't have the correct lapis version on loculus. The config I implemented is some kind of workaround, when we have updated all lapis instances, we can remove it and display the advanced query for all pathogens. 

We could think about restructuring the config for the variant filters. For the baseline filters we have a unified way to configure them. It would be nice to have a similar approach there. 

We could also remove the variantQuery in the future. I think the advanced queries can do everything we want to do with the variantQueries.

### Screenshot

Screenshot 2025-06-11 094817

### DIFF
--- a/website/src/components/genspectrum/AdvancedQueryFilter.tsx
+++ b/website/src/components/genspectrum/AdvancedQueryFilter.tsx
@@ -1,11 +1,14 @@
-import type { ChangeEventHandler } from 'react';
+import type { ChangeEvent } from 'react';
+
+export const advancedQueryUrlParamForVariant = 'advancedQueryVariant';
+export const advancedQueryUrlParam = 'advancedQuery';
 
 export function AdvancedQueryFilter({
     value,
     onInput,
 }: {
     value?: string;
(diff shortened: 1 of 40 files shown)


fengelniederhammer (changes requested):
This has a conceptual issue that we need to discuss first I think: 

image

When I filter for Germany, then I still see the full count. That is expected in a way, since we use the advanced filters for the "variant" part of the filter and not for the baseline, but it looks very surprising. 

Do we need separate advanced query fields for the "dataset filter" and for the "variant filter"?

fengelniederhammer:
Superseded by #735

PR #719, opened 2025-06-11
refactor(website): migrate generic sequencing efforts page to React

### LINKED ISSUE #716: Migrate sequencing efforts page to React

fengelniederhammer:
resolves #716


website/src/components/views/sequencingEfforts/CountStatistics.tsx
```diff
@@ -0,0 +1,46 @@
...
+                        <div className='stat-value text-2xl sm:text-4xl'>{count.toLocaleString('en-us')}</div>
+                        <div className='stat-desc text-wrap'>The total number of sequenced samples</div>
+                    </>
+                )}
+            </div>
+        </div>
+    );
+};
+
+export const CountStatistics = withQueryProvider(CountStatisticsInner);
```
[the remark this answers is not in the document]
anna-parker: I understand that we cannot await in react and need to switch to `useQuery` but I don't fully understand this part

fengelniederhammer:
> I get this warning when I run locally:
> 
> ```
> 10:55:36 [WARN] [vite] Assets in public directory cannot be imported from JavaScript.
> If you intend to import that asset, put the file in the src directory, and use /src/mapData/Democratic Republic of the Congo.topo.json instead of /public/mapData/Democratic Republic of the Congo.topo.json.
> If you intend to use the URL of that asset, use /mapData/Democratic Republic of the Congo.topo.json?url.
> ```

Good catch, the old mechanism didn't really work anymore since it was designed to inspect the map files on server side. Now the component is rendered in the browser, so I think it's better to use a less fancy version that requires a little more manual configuration instead.

website/src/util/getSequencesByLocationMapData.ts
```diff
@@ -1,67 +1,71 @@
...
+function getMapData(mapName: string, mapMetadatum: MapMetadatum): MapData {
     return {
         mapSource: {
             type: 'topojson',
             url: `/mapData/${mapName}.topo.json`,
-            topologyObjectsKey: topologyObjectsKey,
+            topologyObjectsKey: mapMetadatum.topologyObjectsKey,
         },
-        ...mapMetadata[mapName],
+        ...mapMetadatum.additionalInfo,
```
anna-parker, line 68: hey this looks really good - I'm just a bit confused how the map data is now loaded. So I see that baseUrl is actually the page url but the maps are stored in public - how are they available at this url?

fengelniederhammer: Everything that's in `public/` will be bundled to be available at the URL `/`. You can run `npm run build` and see what's in the output (`dist/`). Now to "run the website" from the built output, you basically need a web server that serves `dist/`. Vite makes sure that everything is in the correct location there.

Does that answer you question?

anna-parker: Yes thanks so much! That answers my question! Just to follow up a bit more - the explicit path was only required before then because we were loading it on the server side?

anna-parker: Great thanks for the thorough explanation!

PR #735, opened 2025-06-26
feat(website): add simple text field for advanced queries

### LINKED ISSUE #702: add a simple text field for advanced queries

fengelniederhammer:
resolves #702

### Summary

This adds a text field for an advanced query to every filter section that we have (i.e. dataset filter and variant filter have separate fields). Where necessary, the filters are concatenated with `(query1) and (query2)` before sending them to LAPIS.

Probably we will need some validation of the field. Currently you can easily "break" the page by entering an invalid query and every component will show an error.

I'm also quite sure that we will need some sort of guidance/documentation for users so that they know how to use the field.

You are also likely to see wrong numbers because of https://github.com/GenSpectrum/LAPIS-SILO/issues/830 when playing around with the fields.

### Screenshot

image

## PR Checklist
~~- [ ] All necessary documentation has been adapted.~~

### DIFF
--- a/website/src/components/genspectrum/AdvancedQueryFilter.tsx
+++ b/website/src/components/genspectrum/AdvancedQueryFilter.tsx
@@ -1,12 +1,20 @@
-import type { ChangeEventHandler } from 'react';
+import { type ChangeEvent, type FC, useEffect, useState } from 'react';
 
-export function AdvancedQueryFilter({
-    value,
-    onInput,
-}: {
+export const advancedQueryUrlParamForVariant = 'advancedQueryVariant';
+export const advancedQueryUrlParam = 'advancedQuery';
+
+type AdvancedQueryFilterProps = {
     value?: string;
-    onInput?: ChangeEventHandler<HTMLInputElement>;
-}) {
+    onInput?: (newValue: string | undefined) => void;
+};
+
+export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({ value, onInput }) => {
+    const [inputValue, setInputValue] = useState(value);
+
+    useEffect(() => {
+        setInputValue(value);
+    }, [value]);
+
     return (
         <label className='form-control'>
             <div className='label'>
@@ -15,9 +23,15 @@ export function AdvancedQueryFilter({
             <input
                 className='input input-bordered w-full'
                 placeholder={'Advanced query: A123T & ins_123:TA'}
-                value={value}
-                onInput={onInput}
+                value={inputValue}
+                onInput={(event: ChangeEvent<HTMLInputElement>) => {
+                    const newValue = event.target.value;
+                    setInputValue(newValue === '' ? undefined : newValue);
+                }}
+                onBlur={() => {
+                    onInput?.(inputValue);
+                }}
             ></input>
         </label>
     );
-}
+};
--- a/website/src/components/genspectrum/AdvancedQueryFilter.tsx
+++ b/website/src/components/genspectrum/AdvancedQueryFilter.tsx
@@ -15,6 +15,10 @@ export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({ value, onInp
         setInputValue(value);
     }, [value]);
 
+    if (import.meta.env.PUBLIC_DASHBOARDS_ENVIRONMENT !== 'dashboards-staging') {
+        return null;
+    }
+
     return (
         <label className='form-control'>
             <div className='label'>
--- a/website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
@@ -23,7 +23,7 @@ export function CompareSideBySidePageStateSelector({
 }) {
     const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
     const variantFilterConfig = useMemo(
-        () => makeVariantFilterConfig(view.organismConstants, { enableMutationFilter: true }),
+        () => makeVariantFilterConfig(view.organismConstants),
         [view.organismConstants],
     );
     const [pageState, setPageState] = useState(initialPageState);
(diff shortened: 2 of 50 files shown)


website/src/views/pageStateHandlers/toLapisFilterWithoutVariant.ts
```diff
@@ -1,12 +1,12 @@
 import type { LapisFilter } from '@genspectrum/dashboard-components/util';
 
-import type { Dataset } from '../View.ts';
+import type { DatasetFilter } from '../View.ts';
 
 export function toLapisFilterWithoutVariant(
-    pageState: Dataset,
+    datasetFilter: DatasetFilter,
```
anna-parker, line 6: This change still needs to be made on the GenericAnalyzeSingleVariantPage I think? I am looking at all the parts of the code where toLapisFilterWithoutVariant is used

anna-parker (commented):
It does seem very easy to "break" the page:

image

I think maybe we should hold off on enabling the advanced queries feature until it is able to prevent this

website/src/views/pageStateHandlers/toLapisFilterWithoutVariant.ts
fengelniederhammer: Where exactly do you mean? I didn't find it

(There is
```
    public toLapisFilterWithoutVariant(pageState: DatasetAndVariantData): LapisFilter {
        return toLapisFilterWithoutVariant(pageState.datasetFilter, this.constants.additionalFilters);
    }
```
in case you missed. It that too confusing? Should I revert this change?)

fengelniederhammer:
This also needs https://github.com/GenSpectrum/servers/pull/293

website/src/components/pageStateSelectors/SequencingEffortsPageStateSelector.tsx
```diff
@@ -50,16 +44,21 @@ export function SequencingEffortsPageStateSelector({
                             }));
                         }}
                     />
-                    <VariantSelector
-                        onVariantFilterChange={(newVariantFilter) => {
+                    <GsMutationFilter
```
anna-parker, line 47: the variantSelector (variant queries, advanced queries, lineage filter and GsMutationFilter) has more filter options that the `GsMutationFilter`, I don't understand why this change is being made.

anna-parker:
Questions: 
1. what does the rename of `DASHBOARDS_ENVIRONMENT` to `PUBLIC_DASHBOARDS_ENVIRONMENT` accomplish? In my understanding allowing advancedQueries is now not determined by the config but by hardcoding where it is allowed to be used with 
```
  if (import.meta.env.PUBLIC_DASHBOARDS_ENVIRONMENT !== 'dashboards-staging') {
      return null;
  }
  ```
Couldn't we do this with the `useVariantQuery` and have that set from the config?

**Comment by fengelniederhammer** [comment]
> 1. what does the rename of `DASHBOARDS_ENVIRONMENT` to `PUBLIC_DASHBOARDS_ENVIRONMENT` accomplish?

It makes the env variable accessible in client side code (a vite mechanism)

File: website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx [hunk 3]
```diff
@@ -23,7 +23,7 @@ export function CompareSideBySidePageStateSelector({
 }) {
     const view = useMemo(() => new Routing(organismsConfig), [organismsConfig]).getOrganismView(organismViewKey);
     const variantFilterConfig = useMemo(
-        () => makeVariantFilterConfig(view.organismConstants, { enableMutationFilter: true }),
+        () => makeVariantFilterConfig(view.organismConstants),
```
anna-parker, line 26: why is the `{ enableMutationFilter: true }` removed here and below?

website/src/views/pageStateHandlers/CompareSideBySidePageStateHandler.spec.ts
```diff
@@ -99,7 +99,8 @@ describe('CompareSideBySideStateHandler', () => {
         const url = new URL(
             'http://example.com/covid/compare-side-by-side?' +
                 'columns=3' +
-                '&lineage%241=B.1.1.7&date%241=Last+7+Days' +
+                '&lineage%241=B.1.1.7&advancedQuery%241=country%3DUS&date%241=Last+7+Days' +
+                '&advancedQueryVariant%241=123T' +
```
anna-parker, line 103: why `advancedQueryVariant`? I didn't think we had to split the advanced queries that were related to variants from other advanced queries

website/src/components/pageStateSelectors/CompareSideBySidePageStateSelector.tsx
anna-parker: Ah it seems it is now enabled by default - might be good to add to the PR description as this isn't directly linked to adding advanced queries and it is a larger refactor

website/src/components/genspectrum/AdvancedQueryFilter.tsx
```diff
@@ -1,12 +1,24 @@
...
+export const advancedQueryUrlParam = 'advancedQuery';
+
+type AdvancedQueryFilterProps = {
     value?: string;
-    onInput?: ChangeEventHandler<HTMLInputElement>;
-}) {
+    onInput?: (newValue: string | undefined) => void;
+};
+
+export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({ value, onInput }) => {
```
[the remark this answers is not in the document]
fengelniederhammer: I agree, but I would prefer to do that in a separate step (another subtask of #701)

website/src/components/genspectrum/AdvancedQueryFilter.tsx
```diff
@@ -1,12 +1,24 @@
...
+    const [inputValue, setInputValue] = useState(value);
+
+    useEffect(() => {
+        setInputValue(value);
+    }, [value]);
+
+    if (import.meta.env.PUBLIC_DASHBOARDS_ENVIRONMENT !== 'dashboards-staging') {
+        return null;
+    }
```
fengelniederhammer, line 20: @anna-parker That's what makes it available only on staging

PR #791, opened 2025-09-09
feat(website): expose various component properties on wrappers

fhennig:


website/src/components/genspectrum/GsMutationsOverTime.tsx
```diff
@@ -36,7 +45,11 @@ export const GsMutationsOverTime: FC<GsMutationsOverTimeProps> = ({
                 granularity={granularity}
                 lapisDateField={lapisDateField}
                 displayMutations={displayMutations ? JSON.stringify(displayMutations) : undefined}
+                useNewEndpoint={useNewEndpoint ?? false}
```
fengelniederhammer, line 48: Same trap with boolean attributes: If we write `false` here, then the attibute in the HTML will be `usenewendpoint="false"`, which will result in he value being `true` in the component.
```suggestion
                useNewEndpoint={useNewEndpoint}
```

Using `undefined` for the "false" case works (React will simply omit the attribute, which is what we want)

website/src/components/genspectrum/GsMutationsOverTime.tsx
```diff
@@ -11,6 +16,8 @@ export type GsMutationsOverTimeProps = {
     displayMutations?: string[];
     height?: string;
     pageSizes?: number[];
+    useNewEndpoint?: boolean;
```
fengelniederhammer, line 19:
```suggestion
    useNewEndpoint?: true;
```

website/src/components/genspectrum/GsMutationsOverTime.tsx
```diff
@@ -36,7 +45,11 @@ export const GsMutationsOverTime: FC<GsMutationsOverTimeProps> = ({
                 granularity={granularity}
                 lapisDateField={lapisDateField}
                 displayMutations={displayMutations ? JSON.stringify(displayMutations) : undefined}
+                useNewEndpoint={useNewEndpoint ?? false}
                 pageSizes={JSON.stringify(pageSizes ?? [10, 20, 30, 40, 50])}
+                initialMeanProportionInterval={JSON.stringify(
+                    initialMeanProportionInterval ?? { min: 0.05, max: 0.95 },
```
fengelniederhammer, line 51: Are we intentionally overwriting the default value here? I think it would be better to leave the default value to the component (i.e. not set the attribute if nothing is set in this wrapper here)

fengelniederhammer (changes requested):

website/src/components/genspectrum/GsMutationsOverTime.tsx
fhennig: maybe we could do this?

```
initialMeanProportionInterval={
  initialMeanProportionInterval
    ? JSON.stringify(initialMeanProportionInterval)
    : undefined
}
```

But why is it not done like this for pageSizes?

fengelniederhammer: > maybe we could do this?
> 
> ```
> initialMeanProportionInterval={
>   initialMeanProportionInterval
>     ? JSON.stringify(initialMeanProportionInterval)
>     : undefined
> }
> ```

Looks good to me :+1: 

> But why is it not done like this for pageSizes?

I'm not sure. The default here seems to be the one that the compents already have anyway.

fengelniederhammer: > But you're saying that if I just set it to undefined, it will be omitted?

It might be that it changes with React 19 (since it claims to have better web components support), but for that's how it works.

PR #833, opened 2025-09-23
feat(website): restrict mutation types on wastewater dashboard

### LINKED ISSUE #808: W-ASAP - restrict mutation type inputs
blocked by: https://github.com/GenSpectrum/dashboard-components/pull/980

For the mutation input, restrict the kinds. If we display only nucleotide mutations, only allow those as input.

fhennig:
resolves #808 

### Summary

After it's now possible to restrict the mutation type input, we want only to be able to input _mutations_ (not insertions) and only Nuc or AA, depending on the selected sequence type. This has been implemented here.

### Screenshot

https://github.com/user-attachments/assets/d02d1529-5ed9-4eb6-8bb7-76dddb77c8ba

## PR Checklist

### DIFF
--- a/website/src/components/genspectrum/GsMutationFilter.tsx
+++ b/website/src/components/genspectrum/GsMutationFilter.tsx
@@ -1,5 +1,5 @@
 import '@genspectrum/dashboard-components/components';
-import { gsEventNames } from '@genspectrum/dashboard-components/util';
+import { type MutationType, gsEventNames } from '@genspectrum/dashboard-components/util';
 import { useEffect, useRef } from 'react';
 
 export type MutationFilter = {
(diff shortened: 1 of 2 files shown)


website/src/components/genspectrum/GsMutationFilter.tsx
```diff
@@ -9,13 +9,21 @@ export type MutationFilter = {
     aminoAcidInsertions: string[];
 };
 
+export type MutationType =
+    | 'nucleotideMutations'
+    | 'nucleotideInsertions'
+    | 'aminoAcidMutations'
+    | 'aminoAcidInsertions';
```
fengelniederhammer, line 16: Actually we should also export those in the components rather than duplicating them here. I'd propose to export them as values and as a type:

```
export const mutationType = {
    nucleotideMutations: 'nucleotideMutations',
    ...
}
```

Then use them in the Zod type and also export the Zod type from the components.
(While we're there, we can also do #983 ;) )

website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -161,6 +161,8 @@ function ManualAnalysisFilter({
     pageState: WasapFilter;
     setPageState: (newState: WasapFilter) => void;
 }) {
+    const enabledMutationTypes: MutationType[] =
+        pageState.sequenceType === 'nucleotide' ? ['nucleotideMutations'] : ['aminoAcidMutations'];
```
fengelniederhammer, line 165: Then we could also use the constants here instead of "hardcoding" those strings.

PR #836, opened 2025-09-24
refactor(website): stricter typing for WASAP filter

fhennig:
### Summary

Refactors the `WasapFilter` into a `base` component and into 4 distinct `analysis` types.

In the PageStateSelector, I'm keeping track of all 4 analysis mode settings, so you can go back-and-forth between modes and not lose your settings.

#### Testing

Not sure whether it's better to add tests in the PR as well, or split into a new one. I think I'd be fine with merging as is, and treating testing as a new PR. To test the components, I think I might also do another refactoring and split things into sub-components that can be tested on their own.

### Screenshot
n/a

## PR Checklist

### DIFF
--- a/website/src/views/pageStateHandlers/WasapPageStateHandler.ts
+++ b/website/src/views/pageStateHandlers/WasapPageStateHandler.ts
@@ -37,7 +37,7 @@ export type WasapVariantFilter = {
 
 export type WasapResistanceFilter = {
     mode: 'resistance';
-    sequenceType: 'amino acid';
+    sequenceType: 'amino acid'; // resistance sets are only defined for amino acid mutations
     resistanceSet: ResistanceSetName;
 };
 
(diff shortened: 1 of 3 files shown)


website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -182,3 +203,62 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
...
+    'EG.5',
+    'XBB.1.16',
+    'XBB.1.5',
+    'XBB.2.3',
+    'XBB',
+    'XBB.1.9',
+    'XFG',
+];
+
+export const defaultBaseFilter: WasapBaseFilter = {
```
fengelniederhammer, line 237: This one seems unused?

PR #838, opened 2025-09-25
feat(website): add stats to W-ASAP dashboard

fhennig:
resolves #

## PR Checklist
- [ ] All necessary documentation has been adapted.
- [ ] The implemented feature is covered by an appropriate test.


fhennig:
note to reviewer: It's not perfect to export the fetch fn from the DynamicDateViewer - maybe it'd be better to have some sort of common file where all the fetch functions are? Is that something we should consider doing?

website/src/components/views/wasap/WasapPage.tsx
```diff
@@ -87,6 +89,60 @@ export const WasapPageInner: FC<WasapPageProps> = ({ currentUrl }) => {
...
+                    : isError
+                      ? error.message
+                      : 'The total number of sequenced samples'}
+            </div>
+        </div>
+    );
+};
+
+const formatDate = (d: string) =>
+    new Date(d).toLocaleDateString(undefined, { year: 'numeric', month: 'short', day: 'numeric' });
```
fengelniederhammer, line 114: I think we should fix the locale here so that it looks the same in everyone's browser. What do we want? US locale?
```suggestion
    new Date(d).toLocaleDateString('en-us', { year: 'numeric', month: 'short', day: 'numeric' });
```

Also usually we use this format: `2025-09-19`. The `month: 'short'` also looks nice but I wonder whether we should stick to our usual forma instead?

fengelniederhammer:
> maybe it'd be better to have some sort of common file where all the fetch functions are?

Yes, that would be great! There are a couple of place in the code now where we call LAPIS:
https://github.com/search?q=repo%3AGenSpectrum%2Fdashboards+%2Fsample+language%3ATypeScript&type=code&l=TypeScript

Also there already seems to be a folder that's supposed to be that, I forgot that we have it:
https://github.com/GenSpectrum/dashboards/blob/d9d159f5868f511b76618a3e73d2cfbc476fe62a/website/src/lapis/getTotalCount.ts

It would definitely be good to collect all LAPIS calls in a common spot.

website/src/components/views/wasap/WasapPage.tsx
fhennig: ... done. (see updated screenshot as well)

gordonkoehn (changes requested):
As we are dealing with alignment data, not consensus sequences, we should have a more precise name here.

From the technical sequencing perspective, each sequence in SILO is a **paired-end read**. See here for [context](https://emea.illumina.com/science/technology/next-generation-sequencing/plan-experiments/paired-end-vs-single-read.html). Yet, this is a bit dependent on the technology used; we actually sequence with _Aviti_.

@DrYak suggested **Amplicon Sequences.**, and implies that we merge/paired-end reads or have one-sided reads. 

This is better for mixed audiences, including scientists and non-experts accessing your web platform.

While it is a little less precise on the sequencing technology, it: 
- More accessible to broader audiences, as it emphasizes the biological context (amplicons targeting SARS-CoV-2).
- Aligns with the final data loaded into the database, especially since you merge pairs into consensus sequences.
- Common in wastewater surveillance literature (e.g., ARTIC protocol papers).

With that change, I like the statistics shown a lot.

fhennig:
@gordonkoehn Thanks for the input and explanation! I also had a read about what these terms mean, interesting!

I applied you suggestions in https://github.com/GenSpectrum/dashboards/pull/838/commits/82e763c037e0ecc5016a081c56790900561d0189 and also updated the screenshot above!

gordonkoehn (approved):
Perfect, thanks! Looks Great.

PR #842, opened 2025-09-30
feat(website): define  untracked exclusion list dynamically

### LINKED ISSUE #813: W-ASAP v1
Let me summarise some takeaways from yesterdays's discussion @gordonkoehn, @DrYak and @fhennig - beyond the issues above:

A v1 of the W-ASAP Dashboard we would like to share with the public actively should include the issues linked below, as well as the smaller tasks here:

- [x] merge and deploy this fix: https://github.com/GenSpectrum/dashboard-components/pull/987
- ~~remove 'variant explorer' for now, and change order of other modes: resistance, untracked, manual~~

### Feature overview:
- Explainer Section: we have this data, these modes, for these purposes, these disclaimers
- Resistance Mutations – search known resistance mutations
- Untracked Mutations - search mutations not seen in a set of variants (default: Nextclade Clades)
- Manual Search – search through all mutations by proportion
- Link out single mutations to CovSpectrum on the Plots (i.e. have quick access to CovSpectrum, is this mutation known?)

### LINKED ISSUE #818: W-ASAP untracked mutations: set default list of variants to exclude to set of nextclade variants

fhennig:
resolves #818 

### Summary

- adds a new lapis fetch function that gets the mapping from clade to lineage (with test)
- refactor the UI, so you can explicitly select the Nextstrain clades, or add your own list. The new UI also makes it clear that the list is coming from Nextstrain, and makes it easier to "go back" to that list, in case you edited it.

### Screenshot

https://github.com/user-attachments/assets/64bd35e4-89b3-4ef4-9ecb-2e15858f2267

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
@@ -1,13 +1,17 @@
 import { type SequenceType, mutationType, type MutationType } from '@genspectrum/dashboard-components/util';
+import { type UseQueryResult, useQuery } from '@tanstack/react-query';
 import React, { Fragment, useId, useState } from 'react';
 
 import { ApplyFilterButton } from './ApplyFilterButton';
 import { DynamicDateFilter } from './DynamicDateFilter';
 import { SelectorHeadline } from './SelectorHeadline';
+import { getCladeLineages } from '../../lapis/getCladeLineages';
 import { Inset } from '../../styles/Inset';
 import { wastewaterConfig } from '../../types/wastewaterConfig';
+import { Loading } from '../../util/Loading';
 import { type PageStateHandler } from '../../views/pageStateHandlers/PageStateHandler';
 import {
+    type ExcludeSetName,
     type WasapFilter,
     type WasapAnalysisMode,
     type WasapManualFilter,
@@ -65,6 +69,18 @@ export function WasapPageStateSelector({
                 return { base: baseFilterState, analysis: untrackedFilter };
         }
     }
+
+    // data for the 'untracked' analyis mode - loaded here already so it's available when the mode is selected
+    const cladeLineageQueryResult = useQuery({
+        queryKey: ['cladeLineages'],
+        queryFn: () =>
+            getCladeLineages(
+                wastewaterConfig.wasap.covSpectrum.lapisBaseUrl,
+                wastewaterConfig.wasap.covSpectrum.cladeField,
+                wastewaterConfig.wasap.covSpectrum.lineageField,
+                true,
+            ),
+    });
 
     return (
         <div className='flex flex-col gap-4'>
@@ -141,7 +157,13 @@ export function WasapPageStateSelector({
                                 />
                             );
                         case 'untracked':
-                            return <UntrackedFilter pageState={untrackedFilter} setPageState={setUntrackedFilter} />;
+                            return (
+                                <UntrackedFilter
+                                    pageState={untrackedFilter}
+                                    setPageState={setUntrackedFilter}
+                                    cladeLineageQueryResult={cladeLineageQueryResult}
+                                />
+                            );
                     }
                 })()}
             </Inset>
@@ -199,7 +221,7 @@ function VariantExplorerFilter({
                 onChange={(sequenceType) => setPageState({ ...pageState, sequenceType })}
             />
             <LabeledField label='Variant'>
-                <gs-app lapis={wastewaterConfig.wasap.covSpectrumLapisBaseUrl}>
+                <gs-app lapis={wastewaterConfig.wasap.covSpectrum.lapisBaseUrl}>
                     <GsLineageFilter
                         lapisField='pangoLineage'
                         lapisFilter={{}}
@@ -271,10 +293,15 @@ function ResistanceMutationsFilter({
 function UntrackedFilter({
     pageState,
     setPageState,
+    cladeLineageQueryResult: { isPending, isError, data: cladeLineages },
 }: {
     pageState: WasapUntrackedFilter;
     setPageState: (newState: WasapUntrackedFilter) => void;
+    cladeLineageQueryResult: UseQueryResult<Record<string, string>>;
 }) {
+    const defaultLineages = cladeLineages ? Object.values(cladeLineages) : [];
+    defaultLineages.sort();
+
     return (
         <>
             <SequenceTypeSelector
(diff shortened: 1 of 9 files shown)


website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -271,23 +291,71 @@ function ResistanceMutationsFilter({
...
-                    onChange={(e) => setPageState({ ...pageState, excludeVariants: e.target.value.split(' ') })}
-                />
+                <select
+                    className='select select-bordered'
+                    value={pageState.excludeSet}
+                    onChange={(e) => setPageState({ ...pageState, excludeSet: e.target.value as ExcludeSetName })}
+                >
+                    <option value='nextstrain'>Nextstrain clades</option>
+                    <option value='custom'>custom</option>
+                </select>
```
fengelniederhammer, line 316: Why don't we use the `RadioSelect` select here? This also only has two options. Or do we expect more options in the future?

website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -271,23 +291,71 @@ function ResistanceMutationsFilter({
 function UntrackedFilter({
     pageState,
     setPageState,
+    cladeLineageQueryResult: { isPending, isError, error, data: cladeLineages },
```
fengelniederhammer, line 294: Did you try whether you can simply move the query in here instead of passing the results? Actually `useQuery` should cache the results based on the query key. I think it should work that we don't need to refetch when this component unmounts and is then rendered again.

website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -271,23 +291,71 @@ function ResistanceMutationsFilter({
...
+            {pageState.excludeSet === 'nextstrain' ? (
+                isPending ? (
+                    <Loading />
+                ) : isError ? (
+                    <span>error: {error.message}</span>
+                ) : (
+                    <div className='px-1 py-2 text-sm'>
+                        {defaultLineages.join(', ')}{' '}
+                        <button
+                            className='underline'
```
fengelniederhammer, line 327:
```suggestion
                            className='underline cursor-pointer'
```

website/src/lapis/getCladeLineages.ts
```diff
@@ -0,0 +1,72 @@
+import axios from 'axios';
+import { z } from 'zod';
+
+import { getClientLogger } from '../clientLogger.ts';
+
+const logger = getClientLogger('getCladeLineages');
+
+const aggregatedSchema = z.object({
```
fengelniederhammer, line 8: We should be able to reuse the schema from the other PR once that is merged?

website/src/lapis/getCladeLineages.ts
```diff
@@ -0,0 +1,72 @@
...
+    for (const row of parsedResponse.data.data) {
+        const clade = row[cladeField];
+        const lineage = row[lineageField];
+        const count = row.count;
+
+        if (clade === null || lineage === null) {
+            continue;
+        }
+
+        if (typeof clade !== 'string' || typeof lineage !== 'string' || typeof count !== 'number') {
```
fengelniederhammer, line 61:
```suggestion
        if (typeof clade !== 'string' || typeof lineage !== 'string') {
```

That should have already been validated by the schema.

website/src/lapis/getCladeLineages.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export async function getCladeLineages(
+    baseUrl: string,
+    cladeField: string,
+    lineageField: string,
+): Promise<Record<string, string>> {
+    const url = `${baseUrl.replace(/\/$/, '')}/sample/aggregated`;
+    const body = {
+        fields: [cladeField, lineageField],
+        orderBy: [cladeField],
```
fengelniederhammer, line 31:
```suggestion
        orderBy: [cladeField, { field: 'count', type: 'descending' }],
```

In POST requests, LAPIS can do the sorting already.

Image

website/src/lapis/getCladeLineages.ts
```diff
@@ -0,0 +1,72 @@
...
+            continue;
+        }
+
+        if (typeof clade !== 'string' || typeof lineage !== 'string' || typeof count !== 'number') {
+            throw new Error(`Unexpected row types in clade lineages response: ${JSON.stringify(row)}`);
+        }
+
+        if (!(clade in mapping) || count > mapping[clade].count) {
+            mapping[clade] = { lineage, count };
+        }
```
fengelniederhammer, line 67: Following my comment above: Then we don't need this logic here, do we? We can simply take the first lineage of each clade.

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -192,7 +200,10 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
                 break;
             case 'untracked':
                 setSearchFromString(search, 'sequenceType', analysis.sequenceType);
-                setSearchFromString(search, 'excludeVariants', analysis.excludeVariants?.join('|'));
+                setSearchFromString(search, 'excludeSet', analysis.excludeSet);
+                if (analysis.excludeSet === 'custom') {
+                    setSearchFromString(search, 'excludeVariants', analysis.excludeVariants?.join('|'));
```
fengelniederhammer, line 205: Do we need a check that the URL doesn't get too long? (Same issue that Loculus has)

fengelniederhammer (commented):
Looks good! Just a couple of remarks.

website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
fengelniederhammer: Hm, good point. I don't know from the top of my head. We can leave it as it is :+1:

website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
fhennig: Ok, then I think I'd leave it the way it is for now.

website/src/lapis/getCladeLineages.ts
fhennig: https://github.com/GenSpectrum/dashboards/pull/842/commits/08c5c48aed4681850e191c26a1e3d0e906829a90

website/src/lapis/getCladeLineages.ts
fhennig: https://github.com/GenSpectrum/dashboards/pull/842/commits/08c5c48aed4681850e191c26a1e3d0e906829a90

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
fhennig: Created #848 for this; I think it's a good, valid point and we need to tackle it, but it's not specific to the changes in this PR (the free-text field already existed before).

fengelniederhammer (approved):
LGMT

PR #850, opened 2025-10-02
feat(website): add number fields to the sliders on the W-ASAP dashboard

### LINKED ISSUE #807: W-ASAP - improve sliders
In the 'variant explorer' mode there are two sliders:

Image

But you can't see their current value or the start and end of the slider range. Needs to be improved.

fhennig:
resolves #807 

### Summary

- Adds an input field above the slider

The issue description mentions that the range is not indicated, but by sliding to both ends of the slider, you know the range. The design looks similar to https://components.genspectrum.org/?path=/story/input-number-range-filter--default - which also doesn't have start and end labels.

### Screenshot

image

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
@@ -212,36 +212,24 @@ function VariantExplorerFilter({
                     />
                 </gs-app>
             </LabeledField>
-            <LabeledField label='Min. proportion'>
-                <div className='mb-2 w-full'>
-                    <input
-                        className='w-full'
-                        type='range'
-                        min='0'
-                        max='1'
-                        step='0.01'
-                        value={pageState.minProportion}
-                        onChange={(e) => {
-                            setPageState({ ...pageState, minProportion: parseFloat(e.target.value) });
-                        }}
-                    />
-                </div>
-            </LabeledField>
-            <LabeledField label='Min. count'>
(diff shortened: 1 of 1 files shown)


website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -402,3 +390,46 @@ function LabeledField({ label, children }: { label: string; children: React.Reac
...
+        <LabeledField label={label}>
+            <div className='mb-2 w-full'>
+                <input
+                    className='input input-bordered mb-2 w-full'
+                    type='number'
+                    min={min}
+                    max={max}
+                    step={step}
+                    value={value}
+                    onChange={(e) => onChange(parser(e.target.value))}
```
fengelniederhammer, line 421: I think this also does the job, then we would not need the `parser`. For the min count input, we could maybe add `setPageState({ ...pageState, minCount: Math.round(v) })` (or we don't care that someone can input floats into the text field).
```suggestion
                    onChange={(e) => onChange(Number(e.target.value))}
```

fengelniederhammer, line 421: I think we need a guard here:
```suggestion
                    onChange={(e) => {
                        const value = parser(e.target.value);
                        if (Number.isFinite(value)) {
                            onChange(value);
                        }
                    }}
```

Otherwise typing text into the text field does weird things.

PR #855, opened 2025-10-06
WIP

fhennig:
resolves #

## PR Checklist
- [ ] All necessary documentation has been adapted.
- [ ] The implemented feature is covered by an appropriate test.


fhennig:
@fengelniederhammer maybe have a look.

I noticed that currently, we can only mock a single endpoint once. This PR was supposed to fix that. But it looks like MSW doesn't really want to support this, the way I'm trying to do it.

Everything looks good, except that it seems not possible to prevent MSW from making a real passthrough request if all handlers fail.

In the PR, I made a third request in the dummy test case I used to demo this, and it isn't triggering the `onUnhandledRequest`. I think, if any handler at all matches, it's always considered 'handled' and returning `undefined` basically equals "passthrough".

Any ideas how to solves this in a non-ugly way? Am I 'holding it wrong'?

fengelniederhammer:
> @fengelniederhammer maybe have a look.
> 
> I noticed that currently, we can only mock a single endpoint once. This PR was supposed to fix that. But it looks like MSW doesn't really want to support this, the way I'm trying to do it.
> 
> Everything looks good, except that it seems not possible to prevent MSW from making a real passthrough request if all handlers fail.
> 
> In the PR, I made a third request in the dummy test case I used to demo this, and it isn't triggering the `onUnhandledRequest`. I think, if any handler at all matches, it's always considered 'handled' and returning `undefined` basically equals "passthrough".
> 
> Any ideas how to solves this in a non-ugly way? Am I 'holding it wrong'?

In the components we solved it this way:
https://github.com/GenSpectrum/dashboard-components/blob/25ea169d8830ba365ca5b04423c91b9f68f04c79/components/vitest.setup.ts#L60-L69

The gist is: provide an array of (expected request, expected response). Check if the request matches one of them. If it doesn't, return an error. Does that work in your case here?

PR #863, opened 2025-10-08
feat(website): W-ASAP untracked mode: update values and add help modal

### LINKED ISSUE #858: W-ASAP add info text to 'untracked' mode
explain what we mean with "known variants to exclude"

---

Known variants to exclude: ?

defaults = 9
default = 0.8

Selected variants exclude mutations meeting criteria: present in ≥9 clinical sequences and ≥80% of clinical sequences

---

For every lineage, we look for mutations, that occur at least 9 times and are found in 80% of the samples tagged with that lineage.

Then we take these mutations, and exclude them.

fhennig:
resolves #858 

### Summary

- Add the info modal button and modal. If the button is there, the label gets a margin, so the button doesn't touch the control below it.
- Also updates the values we use for filtering.

### Screenshot

image

image

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
@@ -286,7 +286,17 @@ function KnownVariantsExclusionInfo() {
             </form>
             <h1 className='mb-2 text-xl font-semibold'>How it works</h1>
             <p className='text-gray-700'>
-                For each lineage, mutations appearing at least 9 times and in ≥80% of its samples are excluded.
+                Mutations that are characteristic of selected lineages are excluded based on clinical sequences on{' '}
+                <a className='link' href='https://cov-spectrum.org/'>
+                    CovSpectrum
+                </a>
+                .
+            </p>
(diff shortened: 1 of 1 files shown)


website/src/components/pageStateSelectors/WasapPageStateSelector.tsx
```diff
@@ -277,6 +278,20 @@ function ResistanceMutationsFilter({
     );
 }
 
+function KnownVariantsExclusionInfo() {
+    return (
+        <div className='relative p-8'>
+            <form method='dialog'>
+                <button className='btn btn-sm btn-circle btn-ghost absolute top-2 right-2'>✕</button>
+            </form>
+            <h1 className='mb-2 text-xl font-semibold'>How it works</h1>
```
fengelniederhammer, line 287: (We also have a `ModalHeader` component somewhere, but no need to change this now)

fengelniederhammer (approved):

gordonkoehn (changes requested):
Looks good, perhaps we can be more explicit.

PR #866, opened 2025-10-13
feat(website): guard against too long URLs in the ApplyFilterButton

### LINKED ISSUE #848: W-ASAP - consider maximum URL length
With our free-text fields, the user can potentially enter a lot of stuff, that will then be in the URL. But browsers and web servers have a max length for the URL, so we need to check for that, or we will get broken URLs.

The exact limit depends, we can have a look at Loculus, it seems to be working for them.

A few ideas:
- Check when the user clicks 'apply' if the URL is too long
- Check already when the user enters a lot of stuff in a free text field
- Maybe shorten the parameter names a bit to gain a few characters.

fhennig:
resolves #848 

### Summary

Checks the URL length in the `ApplyFilterButton` and if it's too long, disable the button and display an error message.

Also adds a test file for the button.

### Screenshot

https://github.com/user-attachments/assets/6189ec00-83eb-4a69-a858-47e2e9bfae1b

## PR Checklist


website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    test('should update when state changes from long to short', () => {
+        const longData = 'x'.repeat(2000);
+        const longState: DummyPageState = { data: longData };
+
+        const { container, rerender } = render(
+            <ApplyFilterButton pageStateHandler={handler} newPageState={longState} />,
+        );
+
+        expect(container.querySelector('span.btn-disabled')).toBeInTheDocument();
```
fengelniederhammer: Something is weird here. I was about to suggest using this instead:
```
        expect(getByRole('button', { name: 'Apply filters' })).toBeVisible();
        expect(getByRole('button', { name: 'Apply filters' })).not.toHaveAttribute('href');
```
but then the test fails because it finds multiple buttons if I execute all tests. When I only execute a single test, then it works. Does it maybe not properly clean up after the other tests?

website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+        expect(container.querySelector('a')).not.toBeInTheDocument();
```
fengelniederhammer: I don't really like this assertion because it tests that a CSS class is there. CSS is usually irrelevant for behavior. And maybe we can extract it into a function to reuse it?


```suggestion
        expect(getByRole('button', { name: 'Apply filters' })).toBeVisible();
        expect(getByRole('button', { name: 'Apply filters' })).not.toHaveAttribute('href');
```

website/src/components/pageStateSelectors/ApplyFilterButton.tsx
```diff
@@ -9,9 +13,29 @@ export function ApplyFilterButton<PageState extends object>({
...
+                </a>
+            )}
+            {urlTooLong && (
+                <div className='alert alert-error mt-2'>
+                    <span>
+                        The URL is too long ({fullUrl.length} characters, maximum {MAX_URL_LENGTH}). Please reduce the
+                        amount of data in the filters.
+                    </span>
+                </div>
+            )}
```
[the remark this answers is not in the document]
fhennig: good idea! Sorted: https://github.com/GenSpectrum/dashboards/pull/866/commits/ca474689868119ea372d7f131886ce19387cc060

website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
fhennig: Now I'm just checking for the presence/absence of the href tag, it's just one line, I think I'd not put it into a function. I'm not checking for visibility, do you think that's needed?

website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
fhennig: This is now not relevant anymore with the browser test.

website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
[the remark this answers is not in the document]
fhennig: Yes! Done.

website/src/components/pageStateSelectors/ApplyFilterButton.spec.tsx
fengelniederhammer: I think I would have put the "how to get the button element" into a function. But this is simple enough, we don't necessarily need it.

PR #867, opened 2025-10-13
feat: [W-ASAP] resistance mutations – mature protein annotations

gordonkoehn:
Implements a mutation-level specific, mature protein annotations for the sets of resistance mutations. 

This is very useful to people designing vaccines, as they are about the mature protein products and not the gene positions.

For now, we show a limited set of resistance mutations for which we here once hardcode the translation of these positions of the mutations on the genes, as stored in LAPIS-SILO, to the mature protein products.

### Screenshot
Screenshot 2025-10-13 at 16 38 35

Screenshot 2025-10-13 at 16 38 26

## PR Checklist 
- [x] double-check that mutation translation tables are correct @gordonkoehn 
- [x] check how deletions are handled

### DIFF
--- a/website/src/components/views/wasap/resistanceMutations.ts
+++ b/website/src/components/views/wasap/resistanceMutations.ts
@@ -20,7 +20,7 @@ export const RESISTANCE_MUTATIONS = {
         'ORF1a:M3312I',
         'ORF1a:M3312L',
         'ORF1a:M3312T',
-        'ORF1a:M3312d',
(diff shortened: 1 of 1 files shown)


gordonkoehn:
@fhennig I just created an [EPIC](https://github.com/cbg-ethz/WisePulse/issues/70) where we can track, to remove all these hardcoded annotations. 

With that, I think this is good to merge – please merge at your discretion, you have a better picture of all pending changes.

PR #872, opened 2025-10-14
fix(website): w-ASAP - untracked mutations mode, custom variant input now working again

### LINKED ISSUE #870: W-ASAP - untracked mutations - custom text field erases trailing spaces, making it hard to enter lineages
When you enter something in the text field, you can't enter a space and then the next lineage, the space is instantly erased.

fhennig:
resolves #870 

### Summary
We're using an intermediate string state in the component now, which syncs to the list of variants in a smart way, so we don't lose formatting.

#### Testing

Now I've also started to actually add a test file for one of the components. This is something we want to improve on in the long term anyway. I think it might make sense to split out the sub-components from the big `WasapPageStateSelector` and have individual tests for them, but I didn't want to split out the component in here, I'd rather do a dedicated refactoring PR.

### Screenshot

https://github.com/user-attachments/assets/99096b50-f2df-4e5a-926c-1cea397535cd

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/UntrackedFilter.browser.spec.tsx
+++ b/website/src/components/pageStateSelectors/UntrackedFilter.browser.spec.tsx
@@ -1,10 +1,9 @@
 import { type UseQueryResult } from '@tanstack/react-query';
-import { cleanup, render, waitFor } from '@testing-library/react';
-import { userEvent } from '@testing-library/user-event';
-import '@testing-library/jest-dom/vitest';
-import { afterEach, describe, expect, test, vi } from 'vitest';
+import { describe, expect, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


website/src/components/pageStateSelectors/UntrackedFilter.spec.tsx
```diff
@@ -0,0 +1,208 @@
+import { type UseQueryResult } from '@tanstack/react-query';
+import { cleanup, render, waitFor } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+import '@testing-library/jest-dom/vitest';
```
fengelniederhammer: Why do we need this import?

PR #877, opened 2025-10-15
refactor(website): improve tsconfig and remove React imports

fhennig:
I've had a problem where any TSX syntax was always shown as an error in VSCode, unless I explicitly imported `React`.
I found that when I set an include in the `tsconfig.json`, this was fixed.

We're basing our `tsconfig.json` on the Astro one, and [the docs](https://docs.astro.build/en/guides/typescript/) also recommend to add the include statement, so I did that an the errors disappeared.

Now I've removed all the unnecessary React imports.

(Maybe this should be a `chore`, not a `refactor`; not sure)

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -40,6 +40,7 @@
                 "@testing-library/dom": "^10.4.1",
                 "@testing-library/jest-dom": "^6.8.0",
                 "@testing-library/react": "^16.3.0",
+                "@testing-library/user-event": "^14.6.1",
                 "@types/node": "^24.5.2",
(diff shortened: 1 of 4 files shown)


website/tsconfig.json
```diff
@@ -3,5 +3,7 @@
     "compilerOptions": {
         "jsx": "react-jsx",
         "jsxImportSource": "react"
-    }
+    },
+    "include": [".astro/types.d.ts", "**/*"],
```
fengelniederhammer: Out of interest: Do you know why we need `"**/*"` here?

PR #916, opened 2025-11-13
feat(website): use multi-select lineage filter for untracked mutations

### LINKED ISSUE #843: W-ASAP - untracked mutations - use a UI element that works like the mutation input
Image

fhennig:
resolves #843 

### Summary

In the dashboard-components we now have the `multiSelect` lineage filter! This PR makes use of it.

### Screenshot

https://github.com/user-attachments/assets/818f6843-016a-4bdd-b56e-30c1a96d904a

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/wasap/filters/UntrackedFilter.browser.spec.tsx
+++ b/website/src/components/pageStateSelectors/wasap/filters/UntrackedFilter.browser.spec.tsx
@@ -1,10 +1,14 @@
 import { type UseQueryResult } from '@tanstack/react-query';
+import { userEvent } from '@vitest/browser/context';
 import { describe, expect, vi } from 'vitest';
 import { render } from 'vitest-browser-react';
 
 import { UntrackedFilter } from './UntrackedFilter';
+import type { LapisRouteMocker } from '../../../../../routeMocker';
 import { it } from '../../../../../test-extend';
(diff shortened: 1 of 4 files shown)


website/src/components/pageStateSelectors/wasap/filters/UntrackedFilter.browser.spec.tsx
```diff
@@ -20,177 +24,93 @@ describe('UntrackedFilter - custom variants textarea', () => {
...
-            expect.objectContaining({
+        await vi.waitFor(() => {
+            expect(mockSetPageState).toHaveBeenCalledWith({
+                ...defaultPageState,
                 excludeVariants: ['JN.1', 'KP.2'],
-            }),
-        );
-    });
-
-    it('parses comma-separated variants', async () => {
```
fengelniederhammer: This doesn't work anymore now. Is it important?

fhennig: No, just the LLM thinking it was a good idea to have.

But actually I might go and check what happens when pasting multiple lineages, that might be a useful thing to be able to do. Would also be good to have a test for it.

fhennig: Meh, you can't past multiple things (like you can for the mutation filter), but it's fine for now.

PR #924, opened 2025-11-21
feat(website): add 'Jaccard index' column to the W-ASAP dashboard

### LINKED ISSUE #886: W-ASAP: more "in your face" jaccard index
The recent addition of a jaccard index to convey how much mutations are unique is very useful.

And we would need to find a way to display this information in a much more visible, hard to miss way.

Suggestions:
- colour coding the mutation label on the left axis?
- colour coding the cell frequency table (instead of using blue for all lines, use, e.g, a gradient from red to blue depending on the unicity according to jaccard index?
- display also the jaccard index in the mouse over?
- other suggestions?

fhennig:
resolves #886 

### Summary

Uses the [custom column support](https://github.com/GenSpectrum/dashboard-components/pull/1019) that was added to the mutations-over-time component to display a Jaccard index column when using the variant explorer.

I also refactored the data fetching into a custom hook in a dedicated file.

### Screenshot

image

## PR Checklist
- ~~The implemented feature is covered by an appropriate test.~~ (would be nice to test, but we don't have a framework in place yet)

### DIFF
--- a/website/src/components/views/wasap/useWasapPageData.ts
+++ b/website/src/components/views/wasap/useWasapPageData.ts
@@ -70,22 +70,35 @@ async function fetchMutationSelection(analysis: WasapAnalysisFilter): Promise<Mu
     }
 }
 
-function getDisplayMutationsAndCustomColumns(
-    mutationSelection: MutationSelection | undefined,
-): [string[] | undefined, CustomColumn[] | undefined] {
+/**
+ * The W-ASAP page data consists of the mutations to display in the mutations-over-time component,
+ * and the additional custom columns that might optionally be displayed.
+ *
+ * If displayMutations is undefined, that means that all mutations should be displayed
+ * (That is the default behaviour of the mutations-over-time component).
+ */
+type WasapPageData = {
+    displayMutations?: string[];
+    customColumns?: CustomColumn[];
+};
+
+/**
+ * Turns the internal `MutationSelection` into the easier-to-work-with `WasapPageData`.
+ */
+function wasapPageDataFromMutationSelection(mutationSelection: MutationSelection | undefined): WasapPageData {
     if (mutationSelection === undefined) {
-        return [undefined, undefined];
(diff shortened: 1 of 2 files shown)


website/src/components/views/wasap/useWasapPageData.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export function useWasapPageData(analysis: WasapAnalysisFilter) {
+    const { data, isPending, isError } = useQuery({
+        queryKey: ['wasap', analysis],
+        queryFn: () => fetchMutationSelection(analysis),
+    });
+
+    const [displayMutations, customColumns] = getDisplayMutationsAndCustomColumns(data);
+
+    return { displayMutations, customColumns, isPending, isError };
```
fengelniederhammer: Doesn't matter so much due to how the return value is used, but we lose nice return type of `useQuery` here (i.e. something like: if it's an error, then there must not be data - the caller will not see that. My IDE claims that the return type is `{displayMutations: string[] | undefined, customColumns: {values: Record<string, string | number>, header: string}[] | undefined, isPending: boolean, isError: boolean}`).

I'd suggest to reduce this hook to `return useQuery(...)` and move `getDisplayMutationsAndCustomColumns` into `fetchMutationSelection`.

fengelniederhammer (approved):

website/src/components/views/wasap/useWasapPageData.ts
fhennig: I have modified the function to return the `QueryResult` type directly, by pulling the transforming function into the `queryFn`: https://github.com/GenSpectrum/dashboards/pull/924/commits/3d2416bf6650bd732c91f7ddf7617d1b28701683

I wanted to keep it clean at the call site! Didn't want to have to expose the transform function and internal `MutationSelection` type.

(I also renamed some bit, added some docs)

PR #940, opened 2025-12-02
refactor(website): replace hardcoded wasapPageConfig references with config struct referenced by organism name

fhennig:
This is prep work for being able to more easily add more W-ASAP dashboards for other organisms, like RSV-A.

### Summary

- create a new type `WasapPageConfig` which the `WasapPage` uses and passes down to sub components
- create a new `Wasap.astro` where you only pass in an organism name to get the whole dashboard configured.
- removes old `wasap` section in the `wastewaterConfig`
- create covid W-ASAP config with new type to configure the current dashboard
- make the analysis modes which are enabled configurable, because for RSV-A we will not want all of the ones we currently have
- move all `WasapAnalysisFilter` types also into the `wastewaterConfig` file
- move filter default configurations also into the `WasapPageConfig`

### Screenshot
n/A, looks identical

## PR Checklist
- ~~The implemented feature is covered by an appropriate test.~~ (We already have tests)


fengelniederhammer (approved):
LGTM

fengelniederhammer (approved):

PR #945, opened 2025-12-04
feat(website): add RSV-A W-ASAP dummy page

### LINKED ISSUE #935: Dummy RSV W-ASAP dashboard
Currently we have the Covid wastewater dashboard, but we haven't looked into how we'd set up a second instance for a separate virus yet.

I want to create a RSV dashboard as a dummy for now, using the covid backend, but with it's own config settings and URL etc already in place.

This'll be useful to go through once, and see if there is anything that we have forgotten to make configurable or turn into a generic component.

fhennig:
resolves #935 

### Summary

A lot of code in this PR deals with how to elegantly handle config settings that are absent when a mode is not enabled. This is why there are now a lot more guards etc.

- Changes the `WasapPageConfig` type definition: Now we have specific boolean flags to `enable` a mode. If the flag is set to true, mode specific configs are also added to the type. This allows us to couple settings together, and encode some config setting dependencies in the type system instead of having optional properties that we expect to be there, and then check them individually.
- I've added the menu entries and pages we need, but have not enabled them on prod yet. I've linked GH issues for when we bring stuff live.

### Screenshot

image

## PR Checklist

### DIFF
--- a/website/src/components/views/wasap/wasapPageConfig.ts
+++ b/website/src/components/views/wasap/wasapPageConfig.ts
@@ -47,12 +47,12 @@ export type WasapPageConfigBase = {
  * If the mode is enabled, the type also contains mode dependent settings, like extra fetch settings
  * or mode default configs.
  */
-type AnalysisModeConfigs = ManualAnalyisModeConfig &
+type AnalysisModeConfigs = ManualAnalysisModeConfig &
     VariantAnalysisModeConfig &
     ResistanceAnalysisModeConfig &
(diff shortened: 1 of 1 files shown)


fengelniederhammer (commented):
I see a bunch of requests to https://lapis.pathoplexus.org/rsv-a/sample/aggregated failing with `'pangoLineage' is not a valid sequence filter key.`

fhennig:
I want to rebase onto https://github.com/GenSpectrum/dashboards/pull/953 before merging this PR

PR #953, opened 2025-12-09
fix(website): replace incorrectly hardcoded lapis filter with correctly parameterized filter

fhennig:
### Summary

Another case of an overlooked hardcoded field name.

Found by fengelniederhammer here: https://github.com/GenSpectrum/dashboards/pull/945#pullrequestreview-3556012368

## PR Checklist


fengelniederhammer:
Can't really judge whether it works though without #945, but let's get both merged and then check again. RSV won't be live yet anyway.

PR #961, opened 2025-12-11
feat(website): add time frame selector to W-ASAP variant explorer

### LINKED ISSUE #942: W-ASAP: Variant explorer: Make time frame selectable
Currently we're using the whole dataset of clinical sequences, which goes a few years back (I think). @DrYak mentioned that he ran into the case where there were mutations somewhere far in the past in a now died-out variant, and that caused the Jaccard index for some mutations to be lower than he expected. Maybe we want to change which sequences we use in the selection of clinical sequences, i.e. only use the last year, maybe even also only use Switzerland etc etc. still to be defined.

fhennig:
resolves #942 

### Summary

- Add a new drop-down selector for time frame: "all" (default) "past 3 months", "past 6 months"
- configures the date field as `date` for cov-spectrum and `collectionDateRangeLower` for PPX.
- **Tests**: Add 2 tests in `getMutations.spec.ts` to check that the date parameter is correctly set when querying SILO

### Screenshot

https://github.com/user-attachments/assets/4e695aaa-c4c1-43a5-a210-89b43ff1b334

Documentation:

image

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/wasap/InfoBlocks.tsx
+++ b/website/src/components/pageStateSelectors/wasap/InfoBlocks.tsx
@@ -101,9 +101,8 @@ export function DefineClinicalSignatureInfo() {
             <h2 className='mt-4 mb-2 text-base font-semibold'>Time Frame</h2>
             <p className='text-gray-700'>
                 Time frame for sequence selection. Shorter windows reduce dominance of well-sequenced historical
-                variants, yielding higher Jaccard values for emerging variants with fewer clinical sequences.
-                Use this filter to exclude variants no longer in circulation when assessing significance of
-                current mutations.
+                variants, yielding higher Jaccard values for emerging variants with fewer clinical sequences. Use this
+                filter to exclude variants no longer in circulation when assessing significance of current mutations.
             </p>
         </InfoBlock>
     );
--- a/website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
+++ b/website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
@@ -1,6 +1,10 @@
 import { Inset } from '../../../../styles/Inset';
 import { GsLineageFilter } from '../../../genspectrum/GsLineageFilter';
-import type { VariantTimeFrame, WasapVariantFilter } from '../../../views/wasap/wasapPageConfig';
+import {
+    VARIANT_TIME_FRAME,
+    type VariantTimeFrame,
+    type WasapVariantFilter,
+} from '../../../views/wasap/wasapPageConfig';
 import { SelectorHeadline } from '../../SelectorHeadline';
 import { DefineClinicalSignatureInfo } from '../InfoBlocks';
 import { LabeledField } from '../utils/LabeledField';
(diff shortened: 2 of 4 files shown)


website/src/components/pageStateSelectors/wasap/InfoBlocks.tsx
```diff
@@ -97,6 +97,13 @@ export function DefineClinicalSignatureInfo() {
                 indicate that either not all sequences of the variant have the mutation or not all sequences with the
                 mutation belong to the variant (or both).
             </p>
+
+            <h2 className='mt-4 mb-2 text-base font-semibold'>Time Frame</h2>
+            <p className='text-gray-700'>
+                The time frame of sequences to consider. This is particularily relevant for the Jaccard Index setting.
+                Some mutations might have occurred frequently in other variants, but a relatively long time ago. Chosing
```
gordonkoehn: Chosing -> Choosing

website/src/components/pageStateSelectors/wasap/InfoBlocks.tsx
```diff
@@ -97,6 +97,13 @@ export function DefineClinicalSignatureInfo() {
                 indicate that either not all sequences of the variant have the mutation or not all sequences with the
                 mutation belong to the variant (or both).
             </p>
+
+            <h2 className='mt-4 mb-2 text-base font-semibold'>Time Frame</h2>
+            <p className='text-gray-700'>
+                The time frame of sequences to consider. This is particularily relevant for the Jaccard Index setting.
```
gordonkoehn: I like the text already, but perhaps we can make it more explicit about clinical sequencing efforts, to make clear why this makes sense.

> Time frame for sequence selection. Shorter windows reduce dominance of well-sequenced historical variants, yielding higher Jaccard values for emerging variants with fewer clinical sequences. Use this filter to exclude variants no longer in circulation when assessing significance of current mutations.

What do you think?

gordonkoehn (approved):

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
```diff
@@ -70,6 +70,17 @@ export function VariantExplorerFilter({
...
                 />
+                <LabeledField label='Time frame'>
+                    <select
+                        className='select select-bordered'
+                        value={pageState.timeFrame}
+                        onChange={(e) => setPageState({ ...pageState, timeFrame: e.target.value as VariantTimeFrame })}
+                    >
+                        <option value='all'>All</option>
+                        <option value='6months'>Past 6 months</option>
+                        <option value='3months'>Past 3 months</option>
```
fengelniederhammer: We essentially lose type safety here in the `as VariantTimeFrame` - let's maybe at least use constants so that the values of the options are coupled to the `VariantTimeFrame` type?
```suggestion
                    <select
                        className='select select-bordered'
                        value={pageState.timeFrame}
                        onChange={(e) => setPageState({ ...pageState, timeFrame: e.target.value as VariantTimeFrame })}
                    >
                        <option value={timeFrames.all}>All</option>
                        <option value={timeFrames.6months}>Past 6 months</option>
                        <option value={timeFrames.3months}>Past 3 months</option>
```

and then infer `VariantTimeFrame` from `timeFrames`.

PR #1005, opened 2026-01-27
feat(website): add new collections analysis mode to W-ASAP

### LINKED ISSUE #1006: W-ASAP: cov-spectrum collections over time
As part of #939 we want to start with the existing cov-spectrum collections.

requires the new component: https://github.com/GenSpectrum/dashboard-components/issues/65

fhennig:
resolves #1006

### Summary

Add a new 'Collection' analysis mode, where you can select a [cov-spectrum collection](https://cov-spectrum.org/collections) to analyze, meaning you can see the prevalence of the defined variants over time.

Specific changes:
- New `CollectionAnalysisFilter` component to be used in the `WasapPageStateSelector`.
- `useWasapPageData` was extended to also fetch the collections. During this, the file was also refactored entirely.
- eslint config updated, somehow it was checking node_modules and dist as well?
- the new mode is only enabled in staging.

### Screenshot

https://github.com/user-attachments/assets/3cc41c08-c7c9-4760-a05c-0ca63e7643bb

### Follow-ups 

The generation of the coverage queries still needs to be implemented: https://github.com/GenSpectrum/dashboards/issues/1026 (there is a TODO in the code)

I left some of the refactoring for later, as it's not urgent and I didn't want to put too much into this PR: https://github.com/GenSpectrum/dashboards/issues/1025

Another follow-up: We want to implement a filter for the collections that are shown: https://github.com/GenSpectrum/dashboards/issues/1027

### Useful links

https://github.com/GenSpectrum/cov-spectrum-server/blob/develop/docs/api.md

## PR Checklist

### DIFF
--- a/website/src/components/genspectrum/GsQueriesOverTime.tsx
+++ b/website/src/components/genspectrum/GsQueriesOverTime.tsx
@@ -1,9 +1,10 @@
-import type {
-    CustomColumn,
-    LapisFilter,
-    MeanProportionInterval,
-    QueryDefinition,
-    TemporalGranularity,
+import {
+    views,
+    type CustomColumn,
+    type LapisFilter,
+    type MeanProportionInterval,
+    type QueryDefinition,
+    type TemporalGranularity,
 } from '@genspectrum/dashboard-components/util';
 import { type FC } from 'react';
 
@@ -44,7 +45,7 @@ export const GsQueriesOverTime: FC<GsQueriesOverTimeProps> = ({
                 height={height ? '100%' : undefined}
                 lapisFilter={JSON.stringify(lapisFilter)}
                 queries={JSON.stringify(queries)}
-                views='["grid"]'
+                views={JSON.stringify([views.grid])}
                 granularity={granularity}
                 lapisDateField={lapisDateField}
                 hideGaps={hideGaps}
--- a/website/src/components/genspectrum/GsMutationsOverTime.tsx
+++ b/website/src/components/genspectrum/GsMutationsOverTime.tsx
@@ -1,9 +1,10 @@
-import type {
-    TemporalGranularity,
-    LapisFilter,
-    SequenceType,
-    MeanProportionInterval,
-    CustomColumn,
+import {
+    type TemporalGranularity,
+    type LapisFilter,
+    type SequenceType,
+    type MeanProportionInterval,
+    type CustomColumn,
+    views,
 } from '@genspectrum/dashboard-components/util';
 import { type FC } from 'react';
 
@@ -46,7 +47,7 @@ export const GsMutationsOverTime: FC<GsMutationsOverTimeProps> = ({
                 height={height ? '100%' : undefined}
                 lapisFilter={JSON.stringify(lapisFilter)}
                 sequenceType={sequenceType}
-                views='["grid"]'
+                views={JSON.stringify([views.grid])}
                 granularity={granularity}
                 lapisDateField={lapisDateField}
                 displayMutations={displayMutations ? JSON.stringify(displayMutations) : undefined}


website/src/components/genspectrum/GsQueriesOverTime.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <ComponentWrapper
+            title={'Collection over time' + (collectionTitle ? `: ${collectionTitle}` : '')}
+            height={height}
+        >
+            <gs-queries-over-time
+                width='100%'
+                height={height ? '100%' : undefined}
+                lapisFilter={JSON.stringify(lapisFilter)}
+                queries={JSON.stringify(queries)}
+                views='["grid"]'
```
fengelniederhammer: Nitpick: We could use one of the constants here that the components export.

website/src/components/views/wasap/WasapPage.tsx
```diff
@@ -24,28 +24,30 @@ import {
...
+    isStaging: boolean;
 };
 
-export const WasapPageInner: FC<WasapPageProps> = ({ wastewaterOrganism }) => {
+export const WasapPageInner: FC<WasapPageProps> = ({ wastewaterOrganism, isStaging }) => {
     const config = wastewaterOrganismConfigs[wastewaterOrganism];
     // initialize page state from the URL
     const pageStateHandler = useMemo(() => new WasapPageStateHandler(config), [config]);
 
+    config.collectionAnalysisModeEnabled = isStaging ? true : undefined;
```
fengelniederhammer: It's a bit weird to mutate a global object during rendering. It probably works, but something like

```typescript
const config = useMemo(() => {
    const config = wastewaterOrganismConfigs[wastewaterOrganism]
    return { ...config, collectionAnalysisModeEnabled: isStaging ? true : undefined }
}, [isStating, wastewaterOrganism]);
```

would probably be a bit cleaner

fhennig:
Thanks Fabian! Regarding the query parsing:

So, I've made a mistake, by assuming I can just use any _variant query_ as a LAPIS _advanced query_. They query syntax for both of them has an overlap, but not all variant queries are advanced queries, and vice versa.

I spoke to Chaoran about this, and he said that we should check each variant query and if it isn't a valid advanced query (i.e. we can't use it in LAPIS) we discard it from the collection. This is of course not ideal, but by doing this, we get to use at least some of the cov-spectrum collections. This requires the [LAPIS endpoint to validate queries](https://github.com/GenSpectrum/LAPIS/issues/1518), so we can check if a variant query is valid or not.

Since there is already a TODO related to this blocking LAPIS feature, I'd for now add another TODO in the code that this is a know limitation for now, and then we can tackle both of these things once we have the LAPIS endpoint.

The other thing you mentioned:

> A lot of other collections show this (e.g. collection 211): [screenshot]

Is related to this, in that variants don't perfectly match to queries we can use. I'm only using variants from collections that are defined by a `variantQuery`. Collection 211 doesn't have any variants that are defined in this way, so we end up with an empty list of queries.

I'm not sure how many collections are defined like this, but it is expected that a lot of them won't work. That's also why we want to specifically [filter for wastewater collections](https://github.com/GenSpectrum/dashboards/issues/1027) as a follow-up task.

The specific collection you mentioned defines the variants via the lineage property on the sequences. This can't be applied to wastewater sequences, because the amplicon sequences we have are not assigned to a lineage (they are often too short to confidently assign a linage to them).

--

I hadn't considered it that much before, but now I see that in its current form this feature isn't very useful. I still think it makes sense to maybe postpone the two things above to a future PR, to keep this one kind of small? This one sets up all the 'plumbing' to connect stuff together, in that way I think it's fine to have it as a separate PR. But if you disagree, I can also keep this one open and we do the LAPIS endpoint and the filtering in this PR as well.

(The other stuff you've flagged I will of course fix in this PR!)

Issue #1026, opened 2026-02-05
Wastewater dashboard collections - use LAPIS endpoint to generate coverage queries


fhennig:
Currently, there is only a dummy implementation using a regex; that's not production ready.

~~blocked by: https://github.com/GenSpectrum/LAPIS/issues/1518~~

PR #1050, opened 2026-02-19
feat(website): add new LAPIS endpoint `/query/parse`

fhennig:
part of #1026 (WIP)

### Summary

Adds a new function for query parsing, which uses the new LAPIS endpoint. I've added a test for the API helper, and also a test file specifically for the SiloExpression schema. The schema was initially generated with an LLM.

Also adds integration testing as part of the e2e tests. New command: `npm run e2e:api` which runs the API tests. Four tests were added: smoke test, negative test, multiple queries, nested query.

## PR Checklist


fengelniederhammer (approved):
Nice :)

Issue #938, opened 2025-12-02
Generalized collections


chaoran-chen:
On CoV-Spectrum, a popular feature are the [collections](https://cov-spectrum.org/collections). Any user can create a collection which mainly consists of 

- Title
- Description
- List of variant entry where each only consists of
    - Name
    - A normal variant query or an advanced variant query (the SARS-CoV-2-specific query, not the new advanced query)
    - Description

To create a CoV-Spectrum collection, a user does not need to register but get an admin URL that consists of a key. With the URL, they can edit or delete the collection in the future.

To generalize the feature for whole GenSpectrum / different pathogens, we need:

- Implement an API to create/edit collections in the [backend](https://github.com/GenSpectrum/dashboards/tree/main/backend)
- Define an authentication mechanism (Do we do it the same way as with the CoV-Spectrum collections?)
- Create a web interface to create and maintain collections

fhennig:
Nuclino page for discussion on this issue: https://app.nuclino.com/GenSpectrum/Development/Generalized-collections-e11e1763-e7b4-43ff-b407-db8064356155

fhennig:
issue with follow up issues: https://github.com/GenSpectrum/dashboards/issues/1222

fhennig:
Released on prod!

PR #1081, opened 2026-03-17
feat(website): add API wrapper for collections backend API

fhennig:
This is part of #938 

### Summary

- move existing `backendApi` dir out of the `subscriptions`, into the top level where the other API stuff is also located (`lapis`, `covspectrum`).
- Add new types and functions for the backend API
- Use a common `proxyToBackend` function for subscriptions and collections API
- drop `userId` everywhere, because it's already added later on in the backend proxy

## PR Checklist

### DIFF
--- a/website/src/backendApi/backendProxy.ts
+++ b/website/src/backendApi/backendProxy.ts
@@ -25,13 +25,10 @@ export async function proxyToBackend({ request }: { request: Request }): Promise
 }
 
 /**
- * Like proxyToBackend, but authentication is optional. If the user is logged in, the user ID
- * is forwarded to the backend; if not, the request is proxied without a user ID.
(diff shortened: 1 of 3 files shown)


fengelniederhammer (changes requested):

website/src/backendApi/backendProxy.ts
```diff
@@ -0,0 +1,97 @@
...
+        detail: "You're not authorized to access this resource",
+        status: 401,
+        instance: requestUrl,
+    };
+
+    return Response.json(response, {
+        status: 401,
+        headers: {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            'Content-Type': 'application/json',
```
fengelniederhammer: Claude claims that this header is automatically set by `Response.json`

PR #1155, opened 2026-04-21
fix(website): fix auth secret regenerating on every server restart

### LINKED ISSUE #1145: potential bug: Auth sessions invalidated on every server restart
Claude found this:

---

**File:** `website/src/auth.config.mjs:31`

**Description:**
The Auth.js secret is set to `crypto.randomUUID()`, which generates a new random value every time
 the server process starts. Auth.js uses this secret to sign/encrypt session JWTs. When the
secret changes, all existing session tokens become unverifiable — every user is silently logged
out on each deploy or server restart.

**Impact:**
- Every server restart (deploy, crash, local dev restart) logs out all users
- In production with multiple instances, users logged in on one instance will be logged out when
their next request hits a different instance (each has a different UUID)

**Expected behaviour:**
`AUTH_SECRET` should be a stable, secret value stored in environment configuration — the same
across all instances and restarts.

**Fix:**
Change line 31 to:
```js
secret: process.env.AUTH_SECRET ?? crypto.randomUUID(),
With AUTH_SECRET set in the environment (generate once with openssl rand -base64 32). The
randomUUID() fallback is acceptable for local dev. AUTH_SECRET must be set in staging/production.
```

fhennig:
Remove hardcoded `crypto.randomUUID()` as the auth secret — auth-astro and @auth/core both automatically read AUTH_SECRET from the environment, so the explicit secret field was overriding that and causing all sessions to be invalidated on every restart/deploy.

Add AUTH_SECRET to .env.example with generation instructions.

Fixes #1145

I already rolled out AUTH_SECRET to prod and staging so we can use it. There is also a PR in our servers repo for this.

New auth related tests are added to make sure that a cookie crafted from the value from the environment variable works as a log in token.

I also added a fixture to use an authenticated page directly.

## PR Checklist

### DIFF
--- a/website/.env.example
+++ b/website/.env.example
@@ -6,3 +6,5 @@ LOG_LEVEL=info
 
 GITHUB_CLIENT_ID="Check https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app"
 GITHUB_CLIENT_SECRET="put a secret of you GitHub app here"
+
+AUTH_SECRET="generate with: openssl rand -base64 32"
--- a/website/src/auth.config.mjs
+++ b/website/src/auth.config.mjs
@@ -4,6 +4,7 @@ import { defineConfig } from 'auth-astro';
 import { getGitHubClientId, getGitHubClientSecret } from './config';
 
 export default defineConfig({
+    secret: process.env.AUTH_SECRET,
     callbacks: {
         session: async ({ session, token }) => {
             session.user.id = token.userIdFromProvider;
--- a/website/src/auth.config.mjs
+++ b/website/src/auth.config.mjs
@@ -28,5 +28,4 @@ export default defineConfig({
             clientSecret: getGitHubClientSecret(),
         }),
     ],
-    secret: crypto.randomUUID(),
 });
--- a/website/tests/auth/loginState.spec.ts
+++ b/website/tests/auth/loginState.spec.ts
@@ -1,32 +1,8 @@
-import { encode } from '@auth/core/jwt';
-import { expect, test, type Page } from '@playwright/test';
+import { expect } from '@playwright/test';
 
-const COOKIE_NAME = 'authjs.session-token';
-const TEST_USER_NAME = 'E2E Test User';
+import { test } from '../e2e.fixture.ts';
 
-async function setupAuthCookie(page: Page) {
-    const authSecret = process.env.AUTH_SECRET;
-    if (authSecret === undefined) {
-        throw new Error('AUTH_SECRET environment variable is not set');
-    }
-
-    const token = await encode({
-        token: { userIdFromProvider: 'e2e-test-user', name: TEST_USER_NAME },
-        secret: authSecret,
-        salt: COOKIE_NAME,
-    });
-
-    await page.context().addCookies([
-        {
-            name: COOKIE_NAME,
-            value: token,
-            domain: 'localhost',
-            path: '/',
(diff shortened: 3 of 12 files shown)


website/tests/auth/loginState.spec.ts
fengelniederhammer, line 1: We're testing the prod config in the e2e tests, i.e. those tests here probably won't work...

Should we switch to staging? Actually I think it's better to test what lands on prod.

website/.env.example
```diff
@@ -6,3 +6,5 @@ LOG_LEVEL=info
 
 GITHUB_CLIENT_ID="Check https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app"
 GITHUB_CLIENT_SECRET="put a secret of you GitHub app here"
+
+AUTH_SECRET="generate with: openssl rand -base64 32"
```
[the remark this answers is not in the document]
fhennig: I disagree

website/tests/auth/loginState.spec.ts
fhennig: Hmm, but the tests only depend on the presence of the AUTH_SECRET. And we set that in `.env.e2e` and in the Docker compose file. The config doesn't influence the test.

I think it makes sense to test what lands on prod as well.

fhennig:
Looks like I was wrong to assume that Auth.js automatically finds the AUTH_SECRET? The docs say you should configure it explicitly: https://authjs.dev/reference/core/errors#missingsecret

I'm slightly confused, but have now explicitly configured it. Before I saw this in the website container logs:

```
[auth][error] MissingSecret: Please define a `secret`. Read more at https://errors.authjs.dev#missingsecret
```

even though it is in the env in the container config:

```
  website:
    image: ghcr.io/genspectrum/dashboards/website:${WEBSITE_TAG}
    ...
    environment:
      ...
      AUTH_SECRET: "e2e-test-auth-secret-do-not-use-in-production"
```

website/tests/auth/loginState.spec.ts
fengelniederhammer: AFAIK the whole login button is currently hidden on prod, isn't it?

fhennig:
~~TODO: The login button doesn't show on prod. We decided to enable it for now, in #1164. Once that's on main, I'll rebase the PR and the tests should then work.~~

Edit: rebase done, should be working now 🤞

website/tests/auth/loginState.spec.ts
fhennig: We decided to enable it in #1164

fhennig:
The tests now work. The actual GitHub integration is still to test etc, but that's unrelated to the code we have here, I think we can merge this now @fengelniederhammer

website/src/auth.config.mjs
```diff
@@ -4,6 +4,7 @@ import { defineConfig } from 'auth-astro';
 import { getGitHubClientId, getGitHubClientSecret } from './config';
 
 export default defineConfig({
+    secret: process.env.AUTH_SECRET,
```
fengelniederhammer, line 7: (Super nitpick: Now we could move this to the line where it was before to reduce the diff so that it shows that it just changed the value, not that it was a completely new line. - But feel free to ignore)

fengelniederhammer (approved):

website/src/auth.config.mjs
fhennig: Uhrg, yes you are right ... Eh, not today 😅

PR #1164, opened 2026-04-27
feat(website): enable login button in prod

fhennig:
## Summary

- Removes the `isLoginEnabled()` guard that was hiding the login button in the `dashboards-prod` environment
- Deletes the now-unnecessary `isLoginEnabled` helper from `config.ts`
- Simplifies `CallToAction.astro` and `HamburgerMenu.astro` to render login UI unconditionally

## Testing

I've checked locally that the Login button now shows, when running with DASHBOARDS_ENVIRONMENT `dashboards-prod`

### DIFF
--- a/website/src/config.ts
+++ b/website/src/config.ts
@@ -106,10 +106,6 @@ function getEnvironment() {
     return processEnvOrMetaEnv('DASHBOARDS_ENVIRONMENT', environmentSchema);
 }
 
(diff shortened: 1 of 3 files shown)


fengelniederhammer (approved):
But then let's also check/fix the prod and staging deployments (to make sure that the login buttons work there)

PR #1181, opened 2026-04-28
feat(website): resistance mutation sets as collections

fhennig:
### Summary

- Resistance mutation sets are now sourced from backend collections (IDs 1/2/3) rather than hard-coded arrays in TypeScript
- Collections are fetched server-side in `Wasap.astro` (SSR) and passed as a prop to `WasapPage`, so no client-side fetch or loading state is needed for resistance data
- The config is resolved server-side in `Wasap.astro`, picking prod vs staging collection IDs — staging overrides live in `wastewaterOrganismStagingConfigs`
- `resistanceData.ts` contains the plain async fetch logic; `useResistanceMutationSets` hook deleted
- Prod collections visible at: https://genspectrum.org/api/collections - I already created them so we don't break prod!

### Screenshot
n/a

## PR Checklist


fhennig:
Hey Fabian, the resistance stuff is fetched in the astro frontmatter, I think that's better because it's a simple Docker internal call. But I had it before with a React hook, the change is all in this commit: https://github.com/GenSpectrum/dashboards/pull/1181/commits/2b507415ab037bd979a6e666960035a0133296f6 - Both approaches have pros and cons I'd say, we can also revert this commit if the hook approach is better.

website/src/components/views/wasap/useWasapPageData.ts
```diff
@@ -22,22 +22,32 @@ import { validateGenomeOnly } from '../../../util/siloExpressionUtils';
...
+export function useWasapPageData(
+    config: WasapPageConfig,
+    resistanceMutationsBySet: Record<string, string[]>,
+    analysis: WasapAnalysisFilter,
+) {
     return useQuery({
-        queryKey: ['wasap', analysis],
-        queryFn: () => fetchWasapPageData(config, analysis),
+        queryKey: ['wasap', analysis, resistanceMutationsBySet],
+        queryFn: () => fetchWasapPageData(config, resistanceMutationsBySet, analysis),
```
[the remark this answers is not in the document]
fhennig: no. The object will be stable. I don't think performance will be too crazily impacted. Premature optimization to fiddle with a key here.

website/src/components/views/wasap/useWasapPageData.ts
```diff
@@ -87,14 +97,13 @@ async function fetchVariantModeData(
...
-    }
+function fetchResistanceModeData(
+    displayMutationsBySet: Record<string, string[]>,
+    analysis: WasapResistanceFilter,
+): WasapMutationsData {
     return {
         type: 'mutations',
-        displayMutations:
-            config.resistanceMutationSets.find((set) => set.name === analysis.resistanceSet)?.mutations ?? [],
+        displayMutations: displayMutationsBySet[analysis.resistanceSet ?? ''] ?? [],
```
fengelniederhammer: Can `analysis.resistanceSet` ever be `undefined`?

website/src/components/views/wasap/resistanceData.ts
fengelniederhammer, line 1: Let's maybe add a simple unit test to test the mapping here?

website/src/components/views/wasap/Wasap.astro
```diff
@@ -1,16 +1,34 @@
...
 
 const { wastewaterOrganism } = Astro.props;
-const { name } = wastewaterOrganismConfigs[wastewaterOrganism];
+const config = (isStaging() ? wastewaterOrganismStagingConfigs : wastewaterOrganismConfigs)[wastewaterOrganism];
+const backendService = new BackendService(getBackendHost());
+let resistanceData: ResistanceData = { mutationAnnotations: [], displayMutationsBySet: {} };
+
+try {
+    resistanceData = await fetchResistanceData(config, backendService);
+} catch (error) {
```
fengelniederhammer, line 26: AI: If `fetchResistanceData` fails, the page renders with empty resistance data — the resistance mode dropdown will show zero options, but the user can still select the "resistance" tab and see an empty dropdown. Should we show some kind of user-facing warning in this case, or hide the resistance mode tab entirely when the data is unavailable?

website/src/components/views/wasap/WasapPage.tsx
```diff
@@ -88,6 +76,7 @@ export const WasapPageInner: FC<WasapPageProps> = ({ wastewaterOrganism }) => {
                             initialBaseFilterState={base}
                             initialAnalysisFilterState={analysis}
                             setPageState={setPageState}
+                            resistanceSetNames={Object.keys(displayMutationsBySet)}
                         />
```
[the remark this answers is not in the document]
fhennig: I don't think that's helpful. If the data couldn't be loaded, there is no point in showing set names.

website/src/types/wastewaterConfig.ts
```diff
@@ -160,6 +182,26 @@ export const wastewaterOrganismConfigs: Record<WastewaterOrganismName, WasapPage
...
+    if (!config.resistanceAnalysisModeEnabled) return config;
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const stagingIds: Record<string, number> = { '3CLpro': 1, 'RdRp': 2, 'Spike': 3 };
+    return {
+        ...config,
+        resistanceMutationCollections: config.resistanceMutationCollections.map((set) => ({
+            ...set,
+            collectionId: stagingIds[set.name] ?? set.collectionId,
+        })),
+    };
```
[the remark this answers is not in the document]
fhennig: Well, eventually the IDs might differ. But for now they are identical.

website/src/components/views/wasap/useWasapPageData.ts
fhennig: fixed

website/src/components/views/wasap/Wasap.astro
fhennig: Done.

website/src/components/views/wasap/resistanceData.ts
fhennig: done.

website/src/components/views/wasap/resistanceData.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+                description: 'Main protease inhibitors',
+                aminoAcidMutations: ['S:E484K'],
+            },
+            {
+                name: 'Nirmatrelvir resistant',
+                symbol: 'R',
+                description: 'Main protease inhibitors',
+                aminoAcidMutations: ['S:N501Y'],
+            },
+        ]);
```
fengelniederhammer, line 59: Isn't this wrong? I would expect a single annotation with multiple mutations.

fhennig: No, we are already doing multiple annotations:

image


Maybe the example should be improved.

fhennig: done.

fengelniederhammer: When I log the annotations in the `WasapPage`, then I see a lot of annotations, each with a single mutation:
grafik

Wouldn't it make more sense to group them? Conceptually, each collection should yield a single annotation set with all its mutations.

fhennig: Yes we did that so we can get the "index shift" on them. for A432T this can be 3CLpro:A44T for example (432 -> 44). We wanted to have this (name of the mutation in context of the resistance mutation) individually for each mutation. 

I can see your point though. We had it before in the way you describe it.

fengelniederhammer: As discussed: It makes sense and we'll leave it as it is for now. The filter inside the mutations over time component isn't nice, because it shows a long list of all those mutation annotations, but that's secondary.

fengelniederhammer (approved):

PR #1184, opened 2026-04-29
feat(website): add staging example data seeder

fhennig:
## Summary

- Adds `example-data/` with a Node.js seeder script and Dockerfile that create resistance mutation collections (3CLpro, RdRp, Spike mAb) in the backend
- Seeder runs as a Docker Compose service activated via `--profile staging` — does nothing in normal `docker compose up`
- Idempotent: checks existing collections by name before creating, so re-runs are safe
- No npm dependencies — uses Node 22 built-in `fetch` and `util.parseArgs`
- Persists Postgres data in a named Docker volume across restarts

## CLI usage

The script can also be run by hand against any environment:

```bash
# Local backend
node example-data/seed.js

# Staging or prod via session token (grab __Secure-authjs.session-token from browser DevTools → Application → Cookies)
node example-data/seed.js --url https://staging.genspectrum.org --session-token eyJhbG...

node example-data/seed.js --help  # full options
```

## Docker Compose usage

```bash
BACKEND_TAG=latest WEBSITE_TAG=latest docker compose --profile staging up
```

Note: always pass `--profile staging` to both `up` and `down`, otherwise the seeder container is not included in teardown and may leave stale state.

## Test plan

- [x] `node example-data/seed.js --help` — prints usage, exits 0
- [x] Run with `--profile staging` — seeder starts, waits for backend, creates 3 collections, exits 0
- [x] Run again — all 3 collections are skipped (idempotent)
- [x] Run without `--profile staging` — seeder service does not start
- [x] `docker compose --profile staging down` then `up` — Postgres data is preserved
- [x] Verify via `GET /collections?userId=example-data-seeder&organism=covid`
- [x] Try running the script with a session cookie against staging and see if it still works (maybe modified collection names)

Felix 🤝 🤖 [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/example-data-seeder.yml
+++ b/.github/workflows/example-data-seeder.yml
@@ -1,0 +1,45 @@
+name: Example Data Seeder
+
+on: [push]
+
+env:
+  DOCKER_IMAGE_NAME: ghcr.io/genspectrum/dashboards/example-data-seeder
+
+jobs:
+  dockerImage:
+    name: Build Example Data Seeder Docker Image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Docker metadata
+        id: dockerMetadata
+        uses: docker/metadata-action@v6
+        with:
+          images: ${{ env.DOCKER_IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+
+      - name: Build and push image
+        uses: docker/build-push-action@v7
+        with:
+          context: ./example-data
+          tags: ${{ steps.dockerMetadata.outputs.tags }}
+          cache-from: type=gha,scope=example-data-seeder-${{ github.ref }}
+          cache-to: type=gha,mode=max,scope=example-data-seeder-${{ github.ref }}
+          platforms: ${{ github.ref == 'refs/heads/main' && 'linux/amd64,linux/arm64' || 'linux/amd64' }}
+          push: true
--- a/README.md
+++ b/README.md
@@ -34,11 +34,15 @@ This monorepo contains the following packages:
 - [`backend/`](./backend): 
   The backend for additional features of the dashboard website, currently the notification features.
 
-- [`website/`](./website): 
+- [`website/`](./website):
   The dashboard website: delivery of the (basically static, via Astro) HTML pages with the embedded Dashboard Components
-  (which are included via npm), 
+  (which are included via npm),
   which retrieve data from LAPIS instances directly,
   and some additional client side features accessing the backend.
+
+- [`example-data/`](./example-data):
+  A Node.js seeder script and Dockerfile that populate the backend with example collections (resistance mutation data for 3CLpro, RdRp, and Spike).
+  Activated via the `staging` Docker Compose profile (see below).
 
 
 ## Local setup
@@ -49,3 +53,23 @@ Use Docker Compose to run the dashboards:
 ```bash
 BACKEND_TAG=latest WEBSITE_TAG=latest docker compose up
 ```
+
+### Seeding example data
+
+The `example-data-seeder` service seeds the backend with example collections.
+It only starts when the `staging` Docker Compose profile is active:
+
+```bash
+BACKEND_TAG=latest WEBSITE_TAG=latest docker compose --profile staging up
+```
+
+The seeder is idempotent — running it multiple times will not create duplicate collections.
+Collections are owned by the `example-data-seeder` user and can be inspected via:
+
+```
+GET /collections?userId=example-data-seeder&organism=covid
+```
+
+### Deleting data
+
+`docker compose down` won't delete the postgres volume. Use `... down -v` to also delete the DB volume.
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Check the [Docker compose file](docker-compose.yml) for an example on how to run
 
 Use Docker Compose to run the dashboards:
 ```bash
-BACKEND_TAG=latest WEBSITE_TAG=latest docker compose up
+BACKEND_TAG=latest WEBSITE_TAG=latest SEEDER_TAG=latest docker compose up
 ```
 
 ### Seeding example data
@@ -58,7 +58,7 @@ BACKEND_TAG=latest WEBSITE_TAG=latest docker compose up
 The `example-data-seeder` service seeds the backend with example collections.
 
 ```bash
-BACKEND_TAG=latest WEBSITE_TAG=latest docker compose up
+BACKEND_TAG=latest WEBSITE_TAG=latest SEEDER_TAG=latest docker compose up
 ```
 
 The seeder is idempotent — running it multiple times will not create duplicate collections.
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ This monorepo contains the following packages:
 
 - [`example-data/`](./example-data):
   A Node.js seeder script and Dockerfile that populate the backend with example collections (resistance mutation data for 3CLpro, RdRp, and Spike).
-  Activated via the `staging` Docker Compose profile (see below).
 
 
 ## Local setup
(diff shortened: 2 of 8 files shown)


example-data/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM node:22-alpine
```
fengelniederhammer: Let's use 24 or even 26 (is it already properly released now?), 22 is already quite outdated.

example-data/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Via Docker Compose (staging)
+
+```bash
+BACKEND_TAG=latest WEBSITE_TAG=latest docker compose --profile staging up
+```
+
+## Running locally by hand
+
+Requires Node 18+. No `npm install` needed.
```
fengelniederhammer:
```suggestion
Requires a current version of NodeJS. No `npm install` needed.
```
I'm not sure whether we should mention a specific version here. It's something that is forgotten and gets outdated.

example-data/seed.mjs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    console.log(`\nDone. Created: ${created}, skipped (already exist): ${skipped}.`);
+}
+
+main().catch((err) => {
+    console.error(err);
+    process.exit(1);
+});
```
fengelniederhammer: The rest of the file also uses await, that should actually also work on top level here:


```suggestion
try {
    await main();
} catch (err) {
    console.error(err);
    process.exit(1);
}
```

docker-compose.yml
```diff
@@ -33,3 +33,20 @@ services:
...
     ports:
       - "127.0.0.1:9022:5432"
+    volumes:
+      - database-data:/var/lib/postgresql/data
+
+  example-data-seeder:
+    build:
+      context: ./example-data
+    profiles:
+      - staging
```
fengelniederhammer: I think here we don't need profiles, we can unconditionally use the seeder. This file is only used for local development and e2e tests.

example-data/seed.mjs
```diff
@@ -0,0 +1,291 @@
...
+import { parseArgs } from 'node:util';
+
+const HELP = `\
+Usage: node seed.js [options]
+
+Options:
+  -u, --url <url>              Backend base URL (default: $BACKEND_URL or http://localhost:8080)
+      --user-id <id>           User ID for direct backend access (default: $SEED_USER_ID or example-data-seeder)
+  -t, --session-token <token>  Session token for staging/prod (uses website proxy, injects user from session)
+      --wait                   Retry until backend is ready (auto-enabled when no TTY)
```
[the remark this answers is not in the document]
fengelniederhammer: Maybe we don't even need that option? Will we ever run it against the website proxy?

example-data/Dockerfile
```diff
@@ -0,0 +1,4 @@
...
+WORKDIR /app
+COPY seed.js .
```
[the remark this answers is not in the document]
fengelniederhammer: Good point, probably renaming the file to `.mjs` is the easiest and best option here.

example-data/seed.mjs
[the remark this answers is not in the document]
fhennig: Hmm, I liked having it so I can run it against staging while it's not deployed yet, but that will not matter later on. I also liked it as an example of how one could automatically create collections. But maybe that can also be done in the docs somehow.

I agree that there isn't _really_ a good use case for this in the future, maybe it's best to remove it to keep it simpler. I think I will do that, and add some docs instead somewhere.

docker-compose.yml
```diff
@@ -33,3 +33,17 @@ services:
       POSTGRES_DB: dashboards-backend-db
     ports:
       - "127.0.0.1:9022:5432"
+    volumes:
+      - database-data:/var/lib/postgresql/data
+
+  example-data-seeder:
+    image: ghcr.io/genspectrum/dashboards/example-data-seeder:${SEEDER_TAG}
```
fengelniederhammer, line 40: 🔴 bug: `example-data-seeder` uses `image: ...${SEEDER_TAG}` with no profile gate — it runs unconditionally now. But `e2e.yml` does `docker compose up` with only `WEBSITE_TAG` and `BACKEND_TAG` set. `SEEDER_TAG` is unset → Docker Compose will fail or use empty tag. This is likely why the `tests` (e2e) check is failing. Either set `SEEDER_TAG` in `e2e.yml` (matching the branch tag like the others) + add a `Wait for example-data-seeder Docker image` step, or gate the seeder behind a profile again.

README.md
```diff
@@ -34,12 +34,16 @@ This monorepo contains the following packages:
...
+- [`website/`](./website):
   The dashboard website: delivery of the (basically static, via Astro) HTML pages with the embedded Dashboard Components
-  (which are included via npm), 
+  (which are included via npm),
   which retrieve data from LAPIS instances directly,
   and some additional client side features accessing the backend.
 
+- [`example-data/`](./example-data):
+  A Node.js seeder script and Dockerfile that populate the backend with example collections (resistance mutation data for 3CLpro, RdRp, and Spike).
+  Activated via the `staging` Docker Compose profile (see below).
```
fengelniederhammer: 🟡 stale docs: This still says "Activated via the `staging` Docker Compose profile" but profiles were removed — seeder now runs unconditionally. Should be updated to reflect the new behavior.

README.md
```diff
@@ -49,3 +53,23 @@ Use Docker Compose to run the dashboards:
...
 ```
+
+### Seeding example data
+
+The `example-data-seeder` service seeds the backend with example collections.
+It only starts when the `staging` Docker Compose profile is active:
+
+```bash
+BACKEND_TAG=latest WEBSITE_TAG=latest docker compose --profile staging up
+```
```
fengelniederhammer, line 62: 🟡 stale docs: This section still references `--profile staging` and shows only `BACKEND_TAG` and `WEBSITE_TAG`. Seeder is now unconditional and requires `SEEDER_TAG`. Should be updated, e.g.:

```bash
BACKEND_TAG=latest WEBSITE_TAG=latest SEEDER_TAG=latest docker compose up
```

example-data/seed.mjs
```diff
@@ -0,0 +1,261 @@
...
+
+async function sleep(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function waitForBackend() {
+    for (let attempt = 1; attempt <= RETRY_ATTEMPTS; attempt++) {
+        try {
+            const response = await fetch(
+                `${BACKEND_URL}/collections?userId=${SEED_USER_ID}&organism=covid`,
```
fengelniederhammer, line 183: 🔵 nit: `SEED_USER_ID` not `encodeURIComponent`-wrapped here, but it is in `fetchExistingCollections()` and `createCollection()`. Won't matter for `example-data-seeder` string, but inconsistent.

example-data/seed.mjs
```diff
@@ -0,0 +1,261 @@
...
+const RETRY_DELAY_MS   = 2000;
+const COLLECTIONS_BASE = `${BACKEND_URL}/collections`;
+
+// Converts a genomic mutation code to a mature protein name with the given offset.
+// e.g. matureName("ORF1a:T3284I", "3CLpro", -3263) => "3CLpro:T21I"
+function matureName(mutation, setName, offset) {
+    const mutationPart = mutation.slice(mutation.indexOf(':') + 1);
+    const originalBase = mutationPart[0];
+    const newBase = mutationPart[mutationPart.length - 1];
+    const position = parseInt(mutationPart.match(/\d+/)[0], 10);
```
fengelniederhammer, line 64: 🔵 nit: `.match(/\d+/)` returns `null` on malformed input → crash with cryptic error. Low risk with hardcoded data, but a guard would help if someone adds bad data later.

.github/workflows/example-data-seeder.yml
```diff
@@ -0,0 +1,45 @@
+name: Example Data Seeder
+
+on: [push]
```
fengelniederhammer, line 3: 🔵 nit: No `paths` filter — every push to any branch rebuilds seeder even when `example-data/` is unchanged. Minor CI waste. Other workflows don't filter either so this is consistent, but worth noting.

fengelniederhammer (changes requested):
## AI review: staging example data seeder

Good iteration — session token complexity removed, `.mjs` rename done, Node bumped to 25, CI workflow added. Script is clean and simpler now.

| Severity | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Stale docs | 2 |
| 🔵 Nit | 3 |

**Must fix**: e2e workflow is broken — `SEEDER_TAG` is not set in `e2e.yml`, but the seeder service now runs unconditionally in `docker-compose.yml`. Need to either add `SEEDER_TAG` + wait step to `e2e.yml`, or gate the seeder behind a profile again.

**Should fix**: `README.md` still references `--profile staging` and omits `SEEDER_TAG` — docs are out of sync with the actual compose config after the profile removal.

example-data/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM node:25-alpine
```
fengelniederhammer: Node 25 isn't LTS - let's at least create an issue to update it to 26 as soon as it's released. Then we can stay on 26 for a while.

fhennig: I went down to 24, like our website image. It's also LTS.

fengelniederhammer (approved):

Issue #1188, opened 2026-04-30
The logout button doesn't really log out


fengelniederhammer:
It looks like you're logged out on (/login), but after going to another page, you're logged in again.

Deleting the auth token cookie manually works though. Maybe we're just deleting the wrong cookie?

fhennig:
With better-auth, it works now. So with https://github.com/GenSpectrum/dashboards/pull/1191 it is fixed.

fhennig:
Fixed with recent deployment to prod.

Issue #682, opened 2025-04-25
auth/core updates are broken due to auth-astro


anna-parker:
I ignored these dependabot auth/core updates until we can figure out how to update auth/core with auth-astro as a dependency: https://github.com/nowaythatworked/auth-astro

anna-parker:
See https://github.com/GenSpectrum/dashboards/pull/676 for an example

PR #1191, opened 2026-04-30
feat(website): replace auth-astro with better-auth

### LINKED ISSUE #1189: Replace `auth-astro` with `better-auth`
`auth-astro` (https://github.com/nowaythatworked/auth-astro) is unmaintained and hasn't received updates in a while.

Astro itself has a docs page about [authentication](https://docs.astro.build/en/guides/authentication/#better-auth) where also better auth is mentioned. It looks like a good option, MIT license and has direct GitHub integration support as well.

We want to move to `better-auth`.

---

related: https://github.com/GenSpectrum/dashboards/issues/682

fhennig:
resolves #1189, #682, #1188

## Summary

- Replaces `auth-astro` + `@auth/core` with `better-auth`, running in stateless mode (no database required, JWT-based sessions)
- Removes the custom patch (`patches/auth-astro+4.2.0.patch`) that was needed to fix TypeScript issues in auth-astro

## Changes

- **`src/auth.ts`** (new) — better-auth config with GitHub OAuth provider
- **`src/pages/api/auth/[...all].ts`** (new) — catch-all route handling all auth requests
- **`src/middleware/authMiddleware.ts`** (new) — runs on every request, calls `getSession` once and populates `Astro.locals`
- **`src/backendApi/backendProxy.ts`** — reads `context.locals.user?.githubId` via `APIContext`
- **`src/components/auth/LoginButton.tsx`** — `signIn('github', ...)` → `authClient.signIn.social({ provider: 'github', ... })`
- **`src/auth/logout.ts`** — custom `@auth/core` logout → `auth.api.signOut({ headers, asResponse: true })`
- All pages updated to read from `Astro.locals` instead of calling `auth.api.getSession` directly
- `astro.config.mjs` — removed `auth()` Astro integration

## GitHub user ID handling

better-auth's `session.user.id` is a randomly generated internal ID, not the GitHub numeric user ID. The backend uses the GitHub numeric ID for ownership checks (e.g. who can edit a collection).

The solution: `user.additionalFields` + `mapProfileToUser` in `auth.ts` store the GitHub numeric ID directly on the user object as `githubId` at login time:

```ts
user: {
    additionalFields: {
        githubId: { type: 'string', input: false }
    }
},
socialProviders: {
    github: {
        mapProfileToUser: (profile) => ({ githubId: String(profile.id) })
    }
}
```

`String()` is needed because `profile.id` from GitHub is a number, but the backend stores and compares user IDs as strings. Without the coercion, ownership checks fail due to type mismatch.

This means `session.user.githubId` is the GitHub numeric ID (as a string) everywhere — no secondary lookups needed. The auth middleware sets it on `Astro.locals.user` once per request, and all pages and the backend proxy read it from there.

## E2E test cookie helper

`tests/helpers/auth.ts` provides `createAuthCookies(accountPayload, sessionPayload)` and `setupAuthCookie(page, name)` to craft valid better-auth session cookies for Playwright tests without a real OAuth flow. The cookies are signed/encrypted using the same `AUTH_SECRET` the server uses. `cookieCache` (JWE, 1hr) is enabled so `getSession` can validate the crafted `session_data` cookie without an in-memory store lookup.

## New Cookies

`better-auth` sets 3 cookies. A `session_token`, `session_data` and `account_data`. Session data and account data are JWE, encrypted data.

## Test plan

- [x] Sign in with GitHub works end-to-end
- [x] Logout works
- [x] Ownership checks work correctly (edit/delete only available for your own collections)
- [x] Protected pages (`/collections`, `/subscriptions`) show the login prompt when logged out
- [x] E2E tests pass: `npm run e2e`


website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+            // store the GitHub user ID (i.e. 45882389) in the 'user' object,
+            // which is easy to access from context later on.
+            // the 'id' property cannot be overwritten.
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-conversion -- profile.id is typed as string but GitHub returns a number at runtime; String() ensures it's always stored as a string for consistent ownership checks
+            mapProfileToUser: (profile) => ({ githubId: String(profile.id) }),
+        },
+    },
+    advanced: {
+        trustedProxyHeaders: true,
+        cookiePrefix: 'gen-spectrum',
```
fengelniederhammer: Nitpick: I think usually we use either GenSpectrum or genspectrum

```suggestion
        cookiePrefix: 'genspectrum',
```

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
+import { betterAuth } from 'better-auth';
+
+import { getGitHubClientId, getGitHubClientSecret } from './config';
+
+export const auth = betterAuth({
+    // TODO - maybe we can check again if this is read automatically? Should be, according to the docs.
```
fengelniederhammer: Let's do it?

website/tests/helpers/auth.ts
```diff
@@ -1,28 +1,83 @@
...
-        throw new Error('AUTH_SECRET environment variable is not set');
-    }
+export const E2E_GITHUB_ID = '1234567';
 
-    const token = await encode({
-        token: { userIdFromProvider: userId, name: userId },
-        secret: authSecret,
-        salt: COOKIE_NAME,
-    });
+const SECRET = process.env.AUTH_SECRET ?? '';
```
fengelniederhammer: Do we need to catch the case when this is undefined and throw?

website/tests/helpers/auth.ts
```diff
@@ -1,28 +1,83 @@
...
+    const now = new Date();
+    const expiresAt = new Date(now.getTime() + 7 * 24 * 60 * 60 * 1000);
+    const userId = crypto.randomUUID();
+    const sessionToken = crypto.randomUUID();
+
+    const cookies = await createAuthCookies(
+        {
+            id: crypto.randomUUID(),
+            providerId: 'github',
+            accountId: '1234567',
```
fengelniederhammer:
```suggestion
            accountId: E2E_GITHUB_ID,
```

website/tests/helpers/auth.ts
```diff
@@ -1,28 +1,83 @@
...
+            },
+            user: {
+                id: userId,
+                name,
+                email: 'e2e@test.local',
+                emailVerified: true,
+                image: null,
+                createdAt: now.toISOString(),
+                updatedAt: now.toISOString(),
+                githubId: '1234567',
```
fengelniederhammer:
```suggestion
                githubId: E2E_GITHUB_ID,
```

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+    secret: process.env.AUTH_SECRET,
+    session: {
+        cookieCache: {
+            enabled: true,
+            maxAge: 60 * 60,
+            strategy: 'jwe',
+        },
+    },
```
fengelniederhammer, line 38: 🟡 risk: No `account.storeAccountCookie: true` set. better-auth auto-enables this when no database is provided, so it works today. But if that auto-detection logic ever changes, OAuth account data silently stops persisting. Being explicit is safer for a database-less setup.

```ts
account: {
    storeAccountCookie: true,
},
```

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const auth = betterAuth({
+    // TODO - maybe we can check again if this is read automatically? Should be, according to the docs.
+    secret: process.env.AUTH_SECRET,
+    session: {
+        cookieCache: {
+            enabled: true,
+            maxAge: 60 * 60,
+            strategy: 'jwe',
+        },
```
fengelniederhammer, line 37: 🔴 bug: Missing `refreshCache: true` in `cookieCache`. Without a database, when the 1-hour `maxAge` expires, better-auth tries to refresh from a DB that doesn't exist → user is silently logged out after 1 hour.

The [stateless session docs](https://www.better-auth.com/docs/concepts/session-management#basic-stateless-setup) explicitly show `refreshCache: true` as required for database-less mode. With it, the cookie auto-refreshes at 80% of maxAge (≈48 min), keeping the session alive up to `expiresIn` (default 7 days).

```ts
cookieCache: {
    enabled: true,
    maxAge: 60 * 60,
    strategy: 'jwe',
    refreshCache: true, // ← add this
},
```

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
+import { betterAuth } from 'better-auth';
+
+import { getGitHubClientId, getGitHubClientSecret } from './config';
+
+export const auth = betterAuth({
```
fengelniederhammer, line 8: 🟡 risk: No `baseURL` configured. The [docs](https://www.better-auth.com/docs/reference/options#baseurl) recommend always setting it explicitly:

> Relying on request inference is not recommended. For security and stability, always set `baseURL` explicitly in your config or via the `BETTER_AUTH_URL` environment variable.

The app runs on multiple domains (`genspectrum.org`, `staging.genspectrum.org`, `localhost:4321`). Options:
- Add a `BETTER_AUTH_URL` env var per deployment, or
- Use the dynamic `baseURL` object with `allowedHosts` for multi-domain support

Without it, better-auth infers the URL from the request, which depends on `trustedProxyHeaders` being correctly set and the proxy forwarding correct headers.

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+    // TODO - maybe we can check again if this is read automatically? Should be, according to the docs.
+    secret: process.env.AUTH_SECRET,
+    session: {
```
fengelniederhammer, line 28: 🔵 nit: Consider making `expiresIn` explicit. Defaults to 7 days (`604800`), which is probably fine, but stating it documents the intended session lifetime — especially since `maxAge` (1 hour) is a different concept (cache TTL vs session lifetime) and readers may confuse the two.

```ts
session: {
    expiresIn: 60 * 60 * 24 * 7, // 7 days
    cookieCache: { ... },
},
```

website/src/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+            // the 'id' property cannot be overwritten.
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-conversion -- profile.id is typed as string but GitHub returns a number at runtime; String() ensures it's always stored as a string for consistent ownership checks
+            mapProfileToUser: (profile) => ({ githubId: String(profile.id) }),
+        },
+    },
+    advanced: {
+        trustedProxyHeaders: true,
+        cookiePrefix: 'gen-spectrum',
+    },
+});
```
fengelniederhammer, line 66: 🔵 nit: better-auth logs to console by default. The project uses winston via `getInstanceLogger` everywhere else. Consider wiring them together so auth logs go through the same pipeline:

```ts
import { getInstanceLogger } from './logger';
const logger = getInstanceLogger('auth');

export const auth = betterAuth({
    // ...
    logger: {
        log: (level, message, ...args) => {
            logger[level === 'warn' ? 'warn' : level === 'error' ? 'error' : 'info'](message, ...args);
        },
    },
});
```

Low priority — but without it, auth errors/warnings go to stdout while everything else goes through structured logging.

website/src/env.d.ts
```diff
@@ -1,8 +1,16 @@
 /// <reference path="../.astro/types.d.ts" />
 /// <reference types="astro/client" />
-// declare module 'set-cookie-parser' is added, because tsc checks our dependency auth-astro/server,
-// which uses set-cookie-parser, and it doesn't have types.
-declare module 'set-cookie-parser';
+
+declare namespace App {
+    // Note: 'import {} from ""' syntax does not work in .d.ts files.
```
fengelniederhammer: AI claims that it works syntactically, but that such an import would turn the file into a module which apparently break the global availability of the declared types.

website/src/components/auth/LoginButton.tsx
```diff
@@ -13,13 +14,18 @@ export function LoginButton() {
...
+            .social({
+                provider: 'github',
+                callbackURL: callbackUrlThatDoesNotImmediatelyLogoutAgain,
+            })
+            .catch((error: unknown) => {
+                showErrorToast({
+                    error: error instanceof Error ? error : new Error(String(error)),
+                    logMessage: `Login failed: ${getErrorLogMessage(error)}`,
+                    errorToastMessages: ['Login failed. Please try again.'],
+                });
```
fengelniederhammer, line 27: Preexisting issue, maybe not worth fixing here: I wonder why we didn't put this into a `useMutation` hook? It's an async call, I would expect it to be wrapped for proper handling of errors etc and to follow the pattern.

website/src/components/auth/UserDropdown.astro
```diff
@@ -1,10 +1,8 @@
 ---
-import type { Session } from '@auth/core/types';
-
 import { isStaging } from '../../config';
 
 interface Props {
-    session: Session;
+    session: { user: { name: string | null } };
```
fengelniederhammer: Why don't we reuse the `AuthUser` type somehow that we have in the Astro state?

website/src/components/auth/LoginButton.tsx
```diff
@@ -1,10 +1,11 @@
-import { signIn } from 'auth-astro/client';
+import { createAuthClient } from 'better-auth/client';
```
fengelniederhammer, line 1: The docs mention that there is also a specific import for React:
https://better-auth.com/docs/integrations/astro#create-a-client

Did you try whether it makes a difference? The docs don't say what to expect.

website/src/env.d.ts
```diff
@@ -1,8 +1,16 @@
...
+    // We derive the User type from the auth config so that additionalFields (e.g. githubId) are included.
+    type AuthUser = NonNullable<Awaited<ReturnType<(typeof import('./auth').auth)['api']['getSession']>>>['user'];
+
+    interface Locals {
+        user: AuthUser | null;
```
fengelniederhammer: https://better-auth.com/docs/integrations/astro#astro-locals-types has this:
```suggestion
        user: import("better-auth").User | null;
```

fengelniederhammer (changes requested):
Found a few things, but looks good otherwise 👍

website/src/auth.ts
fhennig: I think this is fine the way it is. we _want_ to pull the URL from the proxy headers, I think?

website/src/components/auth/LoginButton.tsx
fhennig: I didn't try. I think the react client module also includes hooks, but it's not useful for us in that component.

website/src/auth.ts
fengelniederhammer: Probably yes, or would it make sense to set the `BETTER_AUTH_URL` in the deployment?

fhennig: We decided that we want to not set a baseURL, but we want to configure trusted domains. (Where do we pull them from? Hardcoded?)

website/src/env.d.ts
fhennig: We found this to be actually fine. I'll try and maybe improve the documentation.

website/src/auth.ts
fhennig: It was a bit more verbose than I wanted, but I got it set up like this now:

```ts
export const auth = betterAuth({
    logger: {
        level: 'debug',
        log: (level, message) => {
            switch (level) {
                case 'warn':
                    logger.warn(message);
                    break;
                case 'error':
                    logger.error(message);
                    break;
                case 'debug':
                    logger.debug(message);
                    break;
                default:
                    logger.info(message);
            }
        },
    },
```

website/src/components/auth/LoginButton.tsx
fhennig: I made an issue for it: https://github.com/GenSpectrum/dashboards/issues/1209

website/src/env.d.ts
```diff
@@ -1,8 +1,17 @@
...
-declare module 'set-cookie-parser';
+
+declare namespace App {
+    // In auth.ts we define 'additionalFields' for the user type, so we need to define our own AuthUser type
+    // here, based on that (we can't use the plain 'User' from better-auth).  This properly defines the
+    // additional fields.
+    type AuthUser = NonNullable<Awaited<ReturnType<(typeof import('./auth').auth)['api']['getSession']>>>['user'];
+
+    interface Locals {
+        user: AuthUser | null; // we use our own user type, see not above
```
fengelniederhammer, line 11: AI: 🔵 nit: Typo — \"see not above\" → \"see note above\".\n\n```suggestion\n        user: AuthUser | null; // we use our own user type, see note above\n```

website/src/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+                default:
+                    logger.info(message);
+            }
+        },
+    },
+    trustedOrigins: getTrustedOrigins(),
+    session: {
+        expiresIn: 60 * 60 * 24 * 7, // 7 days
+        // The cookie cache in combination with refreshCache: true, means that all the session
+        // information is in a session cookie; the server is completly stateless.
```
fengelniederhammer, line 31: AI: 🔵 nit: Typo — \"completly\" → \"completely\".\n\n```suggestion\n        // information is in a session cookie; the server is completely stateless.\n```

fengelniederhammer (approved):
LGTM 👍

PR #1195, opened 2026-05-05
feat(website): allow restricting which fields are allowed to be referenced in the AdvancedQueryFilter

### LINKED ISSUE #1157: Advanced query filter. supply list of allowed metadata fields
Part 2 of #1094

fhennig:
resolves #1157, #1194

## Summary

Adds an optional `allowedFields` config to the `advancedQuery` baseline filter. When set, the parsed LAPIS expression is walked client-side after successful validation, and any metadata column not in the allowed list causes an error with an informative message. Mutation-only queries (e.g. `A123T`) always pass regardless of `allowedFields`, since they reference sequence positions rather than metadata columns.

- New `AdvancedQueryFilterConfig` type exported from `BaselineSelector.tsx` with optional `allowedFields?: string[]`
- `AdvancedQueryFilter` component accepts and enforces `allowedFields` in its `onSuccess` handler via `extractMetadataFields`
- New `extractMetadataFields` utility in `siloFilterExpression.ts` recursively collects all `column` values from a `SiloFilterExpression` tree
- Three browser spec tests covering: disallowed field → error, allowed field → checkmark + callback, mutation-only → checkmark

The feature is fully opt-in: omitting `allowedFields` preserves existing behaviour (any field is allowed).

## Test plan

- [x] `extractMetadataFields` unit tests in `siloFilterExpression.spec.ts`
- [x] Browser spec tests for the three `allowedFields` scenarios in `AdvancedQueryFilter.browser.spec.tsx`
- [x] Manually tested by temporarily setting `allowedFields: ['host']` on the COVID baseline config — queries on `host` passed, queries on other metadata fields showed the error, mutation queries passed

## PR Checklist

### DIFF
--- a/website/src/components/genspectrum/AdvancedQueryFilter.tsx
+++ b/website/src/components/genspectrum/AdvancedQueryFilter.tsx
@@ -49,7 +49,7 @@ export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({
             const result = results[0];
             if (result.type === 'success') {
                 if (allowedFields !== undefined) {
-                    const usedFields = extractMetadataFields(result.filter);
+                    const usedFields = [...new Set(extractMetadataFields(result.filter))];
                     const disallowed = usedFields.filter((col) => !allowedFields.includes(col));
                     if (disallowed.length > 0) {
                         const listed = disallowed.map((col) => `"${col}"`).join(', ');
--- a/website/src/components/genspectrum/AdvancedQueryFilter.tsx
+++ b/website/src/components/genspectrum/AdvancedQueryFilter.tsx
@@ -37,7 +37,7 @@ export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({
     enabled,
     lapisUrl,
     errorTooltipClass,
-    allowedFields
+    allowedFields,
 }) => {
     const [inputValue, setInputValue] = useState(value);
     const [validationState, setValidationState] = useState<ValidationState>({ type: 'idle' });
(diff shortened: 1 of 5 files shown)


website/src/components/genspectrum/AdvancedQueryFilter.tsx
```diff
@@ -31,6 +39,18 @@ export const AdvancedQueryFilter: FC<AdvancedQueryFilterProps> = ({ value, onInp
...
+                    const usedFields = extractMetadataFields(result.filter);
+                    const disallowed = usedFields.filter((col) => !allowedFields.includes(col));
+                    if (disallowed.length > 0) {
+                        const listed = disallowed.map((col) => `"${col}"`).join(', ');
+                        setValidationState({
+                            type: 'error',
+                            message: `Field ${listed} is not allowed. Allowed fields: ${allowedFields.join(', ')}.`,
+                        });
+                        return;
+                    }
```
fengelniederhammer, line 61: AI: ❓ `allowedFields` is captured in `useMutation`'s `onSuccess` closure. `useMutation` doesn't re-create callbacks when props change — if `allowedFields` ever becomes dynamic (e.g. derived from state), stale closure would use the old list. Fine for now since it comes from static config, but worth noting. Consider using a ref or adding `allowedFields` to the mutation key if this ever becomes dynamic.

website/src/lapis/siloFilterExpression.spec.ts
```diff
@@ -229,7 +233,41 @@ describe('siloFilterExpressionSchema', () => {
...
+                    ],
+                },
+            ],
+        };
+
+        expect(extractMetadataFields(expr)).toEqual(['nextcladePangoLineage', 'country', 'date']);
+    });
+
+    test('returns no columns for mutation-only expressions', () => {
+        const expr: Parameters<typeof extractMetadataFields>[0] = {
```
fengelniederhammer: AI: 🔵 nit: This uses `Parameters<typeof extractMetadataFields>[0]` but the test above uses `SiloFilterExpression` directly (already imported). Inconsistent — prefer `SiloFilterExpression` for clarity.

fengelniederhammer (approved):
AI: Clean, well-scoped feature. Good test coverage. Minor nits inline.

Note: PR body says "resolves #1157, #1194" but #1194 is a dependabot astro bump PR — likely should be just "resolves #1157".

Issue #1190, opened 2026-04-30
collections: display a user name instead of user ID for 'owned by'


fhennig:
Currently, for each collection, we show "owned by 34598223" where 34598223 is the ID of the user that owns the collection. This isn't nice.

I see two options:

A) create a user table (https://github.com/GenSpectrum/dashboards/issues/1179) where we track additional user info such as the name, and then we can display the name.

B) Since we're using GitHub, we could try and directly fetch the GitHub user name and display that, maybe also link to the GitHub profile. (Although it's not straight-forward to fetch user info given the ID)

PR #1201, opened 2026-05-06
feat(backend, website): add users table with GitHub sync and internal IDs, use in website code

### LINKED ISSUE #1179: backend: add user table
We have collections and subscriptions, and they both have owner user IDs. But the ID is just directly taken from the GitHub token we get, we do not track anything else about the user.

This also means that when we want to display a collection and who owns it, we cannot do that, we can only display the user ID.

It would be good to have our own user table.

Then we could also have user profiles where users can explain a bit more about themselves if they want to.

fhennig:
resolves #1179, #1190 

### Summary

- **Backend:** adds a `users` table (BIGSERIAL PK, nullable `github_id`) with upsert via `POST /users/sync` and lookup via `GET /users/{id}`. Migrates `owned_by`/`user_id` FK columns in collections and subscriptions from GitHub ID strings to BIGINT references to the new table.
- **E2E:** updates collection test fixtures to sync a user first and pass the internal ID to collection endpoints.

see also https://github.com/GenSpectrum/dashboards/pull/1202 for more info.

## PR Checklist

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTable.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTable.kt
@@ -11,7 +11,7 @@ const val COLLECTION_TABLE = "collections_table"
 
 object CollectionTable : LongIdTable(COLLECTION_TABLE) {
     val name = text("name")
-    val ownedBy = varchar("owned_by", 255)
+    val ownedBy = long("owned_by")
     val organism = varchar("organism", 255)
     val description = text("description").nullable()
     val createdAt = timestamp("created_at")
@@ -20,7 +20,7 @@ object CollectionTable : LongIdTable(COLLECTION_TABLE) {
 
 class CollectionEntity(id: EntityID<Long>) : LongEntity(id) {
     companion object : LongEntityClass<CollectionEntity>(CollectionTable) {
-        fun findForUser(id: Long, userId: String) = findById(id)
+        fun findForUser(id: Long, userId: Long) = findById(id)
             ?.takeIf { it.ownedBy == userId }
     }
 
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/subscription/SubscriptionModel.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/subscription/SubscriptionModel.kt
@@ -13,18 +13,18 @@ import org.springframework.transaction.annotation.Transactional
 @Service
 @Transactional
 class SubscriptionModel(private val dashboardsConfig: DashboardsConfig) {
-    fun getSubscription(subscriptionId: String, userId: String): Subscription =
+    fun getSubscription(subscriptionId: String, userId: Long): Subscription =
         SubscriptionEntity.findForUser(convertToUuid(subscriptionId), userId)
             ?.toSubscription()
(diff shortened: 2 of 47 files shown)


backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/user/UserModel.kt
```diff
@@ -0,0 +1,35 @@
...
+            existing.toUser()
+        } else {
+            UserEntity.new {
+                githubId = request.githubId
+                name = request.name
+                email = request.email
+                createdAt = now
+                updatedAt = now
+            }.toUser()
+        }
```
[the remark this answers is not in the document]
fhennig: The whole class is annotated as `@Transactional`.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/subscription/SubscriptionModel.kt
```diff
@@ -40,14 +40,14 @@ class SubscriptionModel(private val dashboardsConfig: DashboardsConfig) {
             .toSubscription()
```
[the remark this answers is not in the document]
fhennig: This is a valid point; I think I'll add more explicit checks for the user ID.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTable.kt
```diff
@@ -11,7 +11,7 @@ const val COLLECTION_TABLE = "collections_table"
 
 object CollectionTable : LongIdTable(COLLECTION_TABLE) {
     val name = text("name")
-    val ownedBy = varchar("owned_by", 255)
+    val ownedBy = long("owned_by")
```
fengelniederhammer, line 14: 🟡 risk: Exposed ORM uses plain `long("owned_by")` but the SQL migration adds a real FK constraint (`fk_collections_user`). The ORM layer is unaware of the relationship — no navigation properties, no referential integrity enforcement from Exposed. Same applies to `SubscriptionTable.userId`.

Consider using `reference("owned_by", UserTable)` so the ORM matches the DB schema. This also enables Exposed-level eager/lazy loading if needed later.

fengelniederhammer: AI suggested it, I thought it might be worth looking into it. I'm not sure whether it's really helpful though.

website/tests/collections/collectionForm.spec.ts
```diff
@@ -28,14 +28,27 @@ const SEED_COLLECTION = {
...
+}
+
 test.beforeAll(async ({ request }) => {
-    const response = await request.post(`${BACKEND_URL}/collections?userId=${USER_ID}`, {
+    const syncResponse = await request.post(`${BACKEND_URL}/users/sync`, {
+        data: { githubId: USER_ID, name: 'E2E Test User', email: null },
+    });
+    expect(syncResponse.status()).toBe(200);
+    const syncBody = (await syncResponse.json()) as { id: number };
+    userId = syncBody.id;
```
fengelniederhammer: This is almost worth it's own function, since it appears here and in the `collectionDetail.spec`

fengelniederhammer (approved):

PR #1203, opened 2026-05-07
feat(website): python seeder with pango lineages and test suite

fhennig:
## Summary

- Replaces the JS `seed.mjs` with a unified Python seeder in `collection-seeding/` (renamed from `example-data/`)
- Modular source architecture — new data sources are subclasses of the `Source` ABC; registering a new source is a one-liner in `sources/registry.py`
- Three sources implemented:
  - `covid-resistance-mutations` — port of the original JS resistance mutation data (3CLpro, RdRp, Spike mAb)
  - `covid-pango-lineages` — fetches ~4,976 pango lineage definitions from [corneliusroemer/pango-sequences](https://github.com/corneliusroemer/pango-sequences), one collection per lineage with nucleotide substitutions as variants
  - `covid-pango-lineages-sample` — same as above but limited to 10 lineages, for quick testing (excluded from default run)
- CLI uses `--source NAME` to run a single source; no flag runs all. `--list` prints available sources
- Upserts collections (create or update by name) via `POST /collections` and `PUT /collections/{id}` — no diffing: every existing collection is always updated on each run
- Deletes orphaned collections — each source declares an `owned_tag` (e.g. `#pango-lineage`) which is appended to every collection description. On each run, any existing collection with that tag that is no longer in the source output is deleted via `DELETE /collections/{id}`
- Authenticates via `--api-key` / `$API_KEY`; the system user API key is hardcoded as a dummy value in Docker Compose (both backend and seeder)
- Uses [pixi](https://pixi.sh) for dependency management with a multi-stage Docker build (pixi builder → `python:3.13-slim`)
- Typed throughout with `TypedDict` (`Collection`, `Variant`, `FilterObject`, `ExistingCollection`)
- Supports repeat scheduling via `--repeat-interval-hours` CLI flag or `$REPEAT_INTERVAL_HOURS` env var — if set, the seeder loops with a sleep between runs rather than exiting; Docker Compose runs it every 8 hours with `restart: always`
- 32 tests covering HTTP interactions, mutation name math, collection building, upsert orchestration, and orphan deletion

## Test plan

- [x] `pixi run -e test test` — all 32 tests pass
- [x] `pixi run seed` — seeds all sources against a local backend
- [x] `pixi run seed` again — all collections updated (upsert)
- [x] `pixi run seed-lineages` — seeds all pango lineages
- [x] `pixi run seed-lineages-sample` — seeds 10 lineages quickly


collection-seeding/.dockerignore
fengelniederhammer, line 1: Usually, I did the opposite approach here: I ignored everything and then whitelisted the stuff that I need in the image. It makes sure that new files don't unexpectedly end up in the image (noone ever remembers to update this file).

collection-seeding/api.py
```diff
@@ -0,0 +1,75 @@
...
+                    return
+            except requests.RequestException:
+                pass
+            print(f"Waiting for API... (attempt {attempt}/{attempts})")
+            time.sleep(delay)
+        print(
+            f"API at {self.base_url} did not become ready after {attempts} attempts.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
fengelniederhammer: Let's maybe rather `raise` something? Hard exits in non-top-level scripts are always a bit rude.

collection-seeding/README.md
fengelniederhammer, line 1: Let's maybe add an entry to `docs/arc42/09-architecture-decisions.md`? (Why we need this script and why it ended up being in Python which is a new language to this repo)

collection-seeding/seed.py
```diff
@@ -0,0 +1,143 @@
...
+            total_updated = 0
+            for source in active:
+                c, u = seed_source(client, source)
+                total_created += c
+                total_updated += u
+            if len(active) > 1:
+                print(f"\nTotal — created: {total_created}, updated: {total_updated}.")
+        except Exception as e:
+            print(f"Error: {e}", file=sys.stderr)
+            sys.exit(1)
```
fengelniederhammer: AI: 🟡 risk: `sys.exit(1)` in the loop kills the long-running seeder on any transient API error. When running with `--repeat-interval-hours`, a single failed cycle (e.g. temporary network blip) terminates the process. Consider catching the exception, logging it, and continuing to the next cycle instead of exiting.

collection-seeding/sources/pango_lineages.py
```diff
@@ -0,0 +1,89 @@
...
+
+    name = "covid-pango-lineages"
+
+    def __init__(self, limit: int | None = None):
+        self._limit = limit
+
+    def get_collections(self) -> list[Collection]:
+        print(f"Fetching lineage data from {DATA_URL} ...")
+        response = requests.get(DATA_URL, timeout=60)
+        response.raise_for_status()
```
fengelniederhammer, line 28: AI: 🟡 risk: No error handling for network failure when fetching pango data. `raise_for_status()` throws `requests.HTTPError`, which propagates up to `seed.py`'s catch-all `sys.exit(1)`. In repeat mode, a transient upstream outage kills the entire seeder process. Consider catching and logging so the next cycle can retry.

collection-seeding/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /app
+COPY pixi.toml pixi.lock .
+RUN pixi install --frozen
+
+# Stage 2: slim runtime image — copy only the installed site-packages
+FROM python:3.13-slim AS final
+WORKDIR /app
+COPY --from=builder /app/.pixi/envs/default/lib/python3.13/site-packages \
+                    /usr/local/lib/python3.13/site-packages
+COPY seed.py api.py models.py .
```
fengelniederhammer: AI: 🔵 nit: `COPY seed.py api.py models.py .` — this list needs manual updating whenever a new top-level `.py` file is added. Consider `COPY *.py .` or a `src/` layout to avoid forgetting a file.

collection-seeding/sources/registry.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+from sources.pango_lineages import PangoLineagesSource, PangoLineagesSampleSource
+from sources.resistance_mutations import ResistanceMutationsSource
+from sources import Source
+
+ALL_SOURCES: list[type[Source]] = [
+    ResistanceMutationsSource,
+    PangoLineagesSource,
+    PangoLineagesSampleSource,
```
fengelniederhammer, line 14: AI: ❓ q: `PangoLineagesSampleSource` is in `ALL_SOURCES`, so running without `--source` seeds both the sample (10 lineages) and the full set (~4976 lineages). Since they share the same lineage names, those 10 collections get created by the sample source and then immediately updated by the full source (or vice versa). Is this intentional? Consider excluding the sample from the default run, or documenting that it's only meant for `--source covid-pango-lineages-sample`.

docker-compose.yml
```diff
@@ -39,11 +39,12 @@ services:
...
     image: ghcr.io/genspectrum/dashboards/example-data-seeder:${SEEDER_TAG}
     depends_on:
-      - backend
+      - website
     environment:
-      BACKEND_URL: http://backend:8080
-      SEED_USER_ID: example-data-seeder
-    restart: "no"
+      API_URL: http://website:4321
+      API_KEY: ${SEEDER_API_KEY}
```
fengelniederhammer: We should mention in the README that `SEEDER_API_KEY` needs to be passed here. Doesn't it also need to be passed to the backend somehow?

.github/workflows/example-data-seeder.yml
fengelniederhammer, line 1: Let's also execute the Python tests?

collection-seeding/seed.py
fhennig: Docker would still restart the container.

I think just doing what's suggested here doesn't help us much. We would need some other way to be notified of failures. Looking into the log or into the docker restart count; both not ideal, but fine for now.

collection-seeding/sources/pango_lineages.py
fhennig: yes it's a risk, but I don't think it's a huge risk. And again, Docker will just restart the process. I wouldn't over engineer this now just to mitigate.

docs/arc42/09-architecture-decisions.md
```diff
@@ -29,3 +29,25 @@ We would have needed to host it on one of our servers and configure it properly.
...
+### Context
+
+As GenSpectrum, we want to provide a few collections that we create based on other online resources.
+Ideally, others would maintain their own collections, but for now we want to do it ourselves.
+For example, we create collections for mutations that are relevant for vaccine resistance, based
+on online lists. Or we create collections based on canonical lineage definitions.
+
+Since we want others to also be able to easily generate collections, the code should also serve
+as a kind of reference implementation on how one would generate collections. Therefore we want the
+code to be completely independent from the rest of the codebase, and also understandable to
```
fengelniederhammer: IDE language checks complain:
```suggestion
as a kind of reference implementation on how one would generate collections. Therefore, we want the
code to be completely independent of the rest of the codebase, and also understandable to
```

fengelniederhammer (approved):
👍

PR #1218, opened 2026-05-20
feat(backend, website): add API key authentication for collections API

### LINKED ISSUE #1200: feat: API key authentication for collections API
## Overview

Add support for API key authentication to the collections API, allowing programmatic/script access without a browser-based OAuth session.

## Background

Currently, all authenticated requests go through the Astro frontend server, which validates the user's GitHub OAuth session (via better-auth) and forwards requests to the backend with `?userId=<internalId>`. This works well for browser users but is not suitable for scripts or automation.

## Why not extend better-auth for this?

better-auth does have an `apiKey` plugin, but we are currently using better-auth in a **stateless** mode (session data is stored in a JWE cookie, no database). The `apiKey` plugin requires a database. We intentionally do not want to introduce a separate database just for auth — the backend already has a PostgreSQL instance we can extend instead.

## Design

**One key per user, stored in the backend DB.**

New `api_keys` table (Exposed ORM):
- `id` — Long PK
- `user_id` — FK → `users_table`, unique (enforces one active key per user)
- `key_hash` — SHA-256 of the raw key (raw key is never stored)
- `created_at` — timestamp
- `last_used_at` — timestamp, nullable

**Backend endpoints:**
- `GET /api-keys?userId=` — get current key metadata (404 if none)
- `POST /api-keys?userId=` — generate a new key (409 if one already exists); returns raw key once
- `DELETE /api-keys?userId=` — revoke (deletes the row)
- `POST /internal/api-keys/validate` — called by the proxy; takes raw key, returns userId (safe to leave unprotected as the backend is only reachable within the internal Docker network)

**Proxy change (`authMiddleware.ts`):**
If `Authorization: Bearer <key>` header is present, call `/internal/api-keys/validate` and populate `context.locals.user`. Otherwise fall through to the existing session cookie check. The `backendProxy.ts` `?userId=` forwarding is unchanged.

**Frontend UI:**
Protected page for logged-in users to view key metadata, generate a new key (shown once in a modal), and revoke the current key.

## Future considerations

- Scoped keys (read-only vs read-write)
- Multiple named keys per user

fhennig:
Closes #1200

## Summary

- **DB**: new `api_keys_table` (UUID PK, unique FK to `users_table`, SHA-256 key hash, timestamps)
- **Backend**: `GET/POST/DELETE /api-keys?userId=` for key management; `POST /internal/api-keys/validate` for proxy use; `ConflictException` (409) added to exception handler
- **Proxy**: `authMiddleware` now checks `Authorization: Bearer <key>` first — validates against the backend, populates `gsUserId`, then falls through to the existing session cookie check if absent/invalid; unexpected validation errors are logged
- **Frontend**: `/api-keys` page with a React component to view key metadata, generate a new key (shown once in a modal), and revoke it

## Notes

- The raw key is a 32-byte `SecureRandom` value hex-encoded (64 chars); only its SHA-256 digest is stored

## Tests

**Backend** (`ApiKeyControllerTest`): covers all endpoints — GET/POST/DELETE `/api-keys` and `POST /internal/api-keys/validate`. Cases include: no key returns 404, generated key is 64 hex chars, second key returns 409, revoked key returns 404 on validate, `lastUsedAt` updates on each validation call, two users can hold independent keys.

**Playwright** (`apiKey.spec.ts`): full lifecycle test — navigates to `/api-keys` as an authenticated user, generates a key via the UI, captures it from the show-once modal, creates a collection via `Authorization: Bearer <key>`, verifies the collection is returned, revokes the key in the UI, then confirms a subsequent Bearer request returns 401.

## Screenshots

https://github.com/user-attachments/assets/f36fe82c-5b71-419b-8b9d-7f9ed3543d3a


website/src/pages/api-key/index.astro
fengelniederhammer, line 1: Is there a link to this page anywhere?

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Generates a 32-byte cryptographically random key, stores its SHA-256 digest, and returns the
+     * raw key to the caller. The raw key is never persisted and cannot be retrieved again.
+     */
+    fun generateApiKey(userId: Long): GeneratedApiKey {
+        UserEntity.findById(userId) ?: throw NotFoundException("User $userId not found")
+        if (ApiKeyEntity.findByUserId(userId) != null) {
+            throw ConflictException("An API key already exists for user $userId")
+        }
```
fengelniederhammer, line 31: 🟡 risk: TOCTOU race between `findByUserId` check (L27) and `ApiKeyEntity.new` (L34). Two concurrent requests can both pass the check. The `@Transactional` default isolation (READ COMMITTED on Postgres) does not prevent this — both transactions can read "no key exists" concurrently. The DB unique index catches it, but throws a raw constraint violation (500 instead of the clean 409 `ConflictException`).

Two fix options:
1. **Catch the constraint violation** (`DataIntegrityViolationException`) and map to `ConflictException` — simpler, recommended
2. **Set `@Transactional(isolation = Isolation.SERIALIZABLE)`** on this method — prevents the race at DB level but adds overhead and needs retry logic for serialization failures

website/src/components/api-key/ApiKeyManager.tsx
```diff
@@ -0,0 +1,142 @@
...
+                                        revokeMutation.mutate();
+                                    }
+                                }}
+                                disabled={revokeMutation.isPending}
+                            >
+                                {revokeMutation.isPending ? 'Revoking…' : 'Revoke key'}
+                            </button>
+                            {revokeMutation.isError && (
+                                <p className='mt-2 text-sm text-red-600'>Failed to revoke key. Please try again.</p>
+                            )}
```
fengelniederhammer: 🔵 nit: There's a reusable DaisyUI `Modal` component at `styles/containers/Modal.tsx` with `useModalRef`, `ModalHeader`, `ModalContent` — used across subscriptions, error reports, and filters. It handles backdrop close and native `<dialog>` behavior (Escape key, focus). Consider using it instead of the custom div-based overlay.

backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/ApiKeyControllerTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    @param:Autowired private val apiKeyClient: ApiKeyClient,
+    @param:Autowired private val usersClient: UsersClient,
+) {
+    // --- GET /api-keys ---
+
+    @Test
+    fun `WHEN no key exists THEN GET returns 404`() {
+        val userId = usersClient.createUser()
+
+        apiKeyClient.getApiKeyRaw(userId)
```
fengelniederhammer, line 29: 🔵 nit: Test names use `WHEN...THEN` format, but the project convention (per `backend/AGENTS.md` and all other controller tests) is `` `GIVEN ... WHEN ... THEN ...` `` backtick format. E.g. this could be `` `GIVEN no key exists WHEN getting API key THEN returns 404` ``.

website/src/components/api-key/ApiKeyManager.tsx
```diff
@@ -0,0 +1,142 @@
...
+                <div
+                    role='dialog'
+                    aria-modal='true'
+                    aria-label='Your new API key'
+                    className='fixed inset-0 z-50 flex items-center justify-center bg-black/50'
+                >
+                    <div className='w-full max-w-lg rounded-lg bg-white p-6 shadow-xl'>
+                        <h2 className='mb-2 text-lg font-semibold'>Your new API key</h2>
+                        <p className='mb-3 text-sm text-gray-600'>Copy this key now — it will not be shown again.</p>
+                        <code className='mb-4 block rounded bg-gray-100 p-3 text-sm break-all'>{generatedKey}</code>
```
fengelniederhammer: ❓ question: The modal shows a 64-char hex key with "Copy this key now" message but no clipboard copy button. Manual selection of a long hex string is error-prone. Consider adding a "Copy to clipboard" button (`navigator.clipboard.writeText`).

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private fun generateRawKey(): String {
+        val bytes = ByteArray(32)
+        secureRandom.nextBytes(bytes)
+        return bytes.joinToString("") { "%02x".format(it) }
+    }
+
+    private fun sha256(input: String): String {
+        val bytes = MessageDigest.getInstance("SHA-256").digest(input.toByteArray())
```
fengelniederhammer: 🔵 nit: `toByteArray()` without explicit charset. Should be `toByteArray(Charsets.UTF_8)` for deterministic hashing across environments. Kotlin defaults to UTF-8 on most JVMs, but being explicit is safer for a security-sensitive function.

website/src/components/api-key/ApiKeyManager.tsx
fengelniederhammer: Or maybe don't even make it a modal? Wheneven I've seen such a thing somewhere else, it was usually just a line on the regular page. I think I'd prefer that.

website/src/pages/api-key/index.astro
fhennig: no.

fengelniederhammer: Should we add one in the user menu on the top right?

backend/src/main/resources/db/migration/V1.4__add_api_keys_table.sql
```diff
@@ -0,0 +1,9 @@
+create table api_keys_table (
+    id uuid primary key default gen_random_uuid(),
+    user_id bigint not null unique references users_table(id),
```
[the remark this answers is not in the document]
fhennig: It's working for me though. But I guess being explicit can't hurt.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -0,0 +1,68 @@
...
+            ?: throw NotFoundException("No API key found for user $userId")
+        entity.delete()
+    }
+
+    /** Hashes [key], looks it up, updates [ApiKeyTable.lastUsedAt], and returns the owner's user ID. */
+    fun validateApiKey(key: String): Long {
+        val entity = ApiKeyEntity.findByKeyHash(sha256(key))
+            ?: throw NotFoundException("Invalid API key")
+        entity.lastUsedAt = now()
+        return entity.userId
```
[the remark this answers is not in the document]
fhennig: IMO I would favor KISS over any modifications here.

Realistically, we won't get enough traffic to have to think about this. I'd rather introduce API rate limits to reduce access. Also we don't expect the same API key to be used concurrently.

@fengelniederhammer what do you think?

website/src/components/auth/UserDropdown.astro
```diff
@@ -24,6 +24,9 @@ const showSubscriptions = isStaging();
                 </li>
             )
         }
+        <li>
+            <a href='/api-key'>API Key</a>
+        </li>
```
fengelniederhammer, line 29: Let's also add it to the hamburger menu:

Image

website/tests/api-key/ApiKeyPage.ts
```diff
@@ -0,0 +1,29 @@
...
+    public generateButton() {
+        return this.page.getByRole('button', { name: 'Generate API key' });
+    }
+
+    public revokeButton() {
+        return this.page.getByRole('button', { name: 'Revoke key' });
+    }
+
+    public generatedKey() {
+        return this.page.locator('div.bg-gray-100 code');
```
fengelniederhammer: Let's please not select by CSS classes. If we need a better identifier than what is visible as text / role, we can add a testid or similar.

website/tests/api-key/apiKey.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    organism: ORGANISM,
+    description: 'Created via Bearer token auth by E2E tests — safe to delete',
+    variants: [],
+};
+
+test.describe('API key management', () => {
+    let userId: number;
+
+    test.beforeAll(async ({ request }) => {
+        userId = await syncUser(request, E2E_GITHUB_ID);
```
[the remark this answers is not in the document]
fhennig: no, we can do this at some point in the future if we run into problems.

website/src/components/auth/UserDropdown.astro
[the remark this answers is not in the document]
fhennig: updated the PR desc.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -0,0 +1,74 @@
...
+        val now = now()
+
+        try {
+            ApiKeyEntity.new {
+                this.userId = userId
+                this.keyHash = sha256(rawKey)
+                this.createdAt = now
+                this.lastUsedAt = null
+            }
+        } catch (e: DataIntegrityViolationException) {
```
fengelniederhammer, line 43: `e` is unused - let's maybe log it? (Or at least something - we need to be careful to not log any API keys here.)

fhennig: Do we really need to log this? It's a user error and they get the conflict exception, I think that's fine.

website/tests/api-key/ApiKeyPage.ts
fhennig: I've now added a test ID

website/src/components/auth/UserDropdown.astro
fhennig: done.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
fengelniederhammer: As discussed, fine as is it.

fengelniederhammer (approved):

PR #1220, opened 2026-05-20
feat(backend): add system user auto-creation from application YAML

fhennig:
For our staging environment, we want to initialize a system user with a known API key, that we can use right away from our collection seeding scripts.

This feature allows us to configure that system user in the `applications.yaml`:

```yaml
dashboards:
  systemUser:
    githubId: "218605180"
    name: "Genspectrum Bot"
    email: null
    apiKey: "${SYSTEM_USER_API_KEY}"
  organisms:
    ...
```

And then set the env var (`SYSTEM_USER_API_KEY`).

In our deployment scripts we can then have a shared secret between the backend and the seeder container.

## Summary

- Adds `SystemUserConfig` data class and optional `systemUser` field to `DashboardsConfig`
- Adds `upsertApiKey(userId, rawKey)` to `ApiKeyModel` — inserts on first run, rotates on key change, no-ops if hash is unchanged
- New `SystemUserInitializer` `ApplicationRunner` creates/updates the configured bot account and upserts its API key at startup

## Test plan

- [x] 4 integration tests in `SystemUserInitializerTest` cover: key creation + user existence, idempotency, key rotation (old key rejected / new key validates), `lastUsedAt` not reset on same-key re-run
- [x] Add a `systemUser` block to a local YAML profile, start the backend, verify user appears at `GET /users/{id}` and key validates at `POST /internal/api-keys/validate`
- [x] Restart with identical config → same user, same key, no errors
- [x] Restart with changed `apiKey` → old key rejected, new key validates

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
@@ -4,7 +4,7 @@ import org.genspectrum.dashboardsbackend.controller.BadRequestException
 import org.springframework.boot.context.properties.ConfigurationProperties
 
 @ConfigurationProperties(prefix = "dashboards")
-data class DashboardsConfig(val organisms: Map<String, OrganismConfig>) {
+data class DashboardsConfig(val organisms: Map<String, OrganismConfig>, val systemUser: SystemUserConfig? = null) {
     fun getOrganismConfig(organism: String) = organisms[organism]
         ?: throw IllegalArgumentException("No configuration found for organism $organism")
 
@@ -35,3 +35,10 @@ data class LapisConfig(
 )
 
 data class ExternalNavigationLink(val url: String, val label: String, val menuIcon: String, val description: String)
+
+data class SystemUserConfig(
+    val githubId: String,
+    val name: String,
+    val email: String? = null,
+    val apiKey: String? = null,
+)
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
@@ -41,4 +41,8 @@ data class SystemUserConfig(
     val name: String,
     val email: String? = null,
     val apiKey: String? = null,
-)
+) {
+    init {
+        require(apiKey == null || apiKey.length >= 32) { "systemUser.apiKey must be at least 32 characters" }
+    }
+}
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
@@ -45,4 +45,7 @@ data class SystemUserConfig(
     init {
         require(apiKey == null || apiKey.length >= 32) { "systemUser.apiKey must be at least 32 characters" }
     }
+
+    override fun toString() =
+        "SystemUserConfig(githubId=$githubId, name=$name, email=$email, apiKey=${if (apiKey != null) "***" else "null"})"
 }
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializer.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializer.kt
@@ -21,7 +21,7 @@ class SystemUserInitializer(
         val user = userModel.syncUser(
             UserSyncRequest(githubId = config.githubId, name = config.name, email = config.email),
         )
-        log.info { "System user ready: id=${user.id}, githubId=${config.githubId}, name=${config.name}" }
+        log.info { "System user ready: id=${user.id}, $config" }
         config.apiKey?.let {
             apiKeyModel.upsertApiKey(userId = user.id, rawKey = it)
             log.info { "System user API key upserted for userId=${user.id}" }
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializer.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializer.kt
@@ -1,0 +1,30 @@
+package org.genspectrum.dashboardsbackend.config
+
+import mu.KotlinLogging
+import org.genspectrum.dashboardsbackend.api.UserSyncRequest
+import org.genspectrum.dashboardsbackend.model.apikey.ApiKeyModel
+import org.genspectrum.dashboardsbackend.model.user.UserModel
+import org.springframework.boot.ApplicationArguments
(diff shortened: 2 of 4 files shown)


backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/DashboardsConfig.kt
```diff
@@ -35,3 +35,10 @@ data class LapisConfig(
 )
 
 data class ExternalNavigationLink(val url: String, val label: String, val menuIcon: String, val description: String)
+
+data class SystemUserConfig(
+    val githubId: String,
+    val name: String,
+    val email: String? = null,
+    val apiKey: String? = null,
+)
```
fengelniederhammer: 🟡 risk: `SystemUserConfig` is a data class, so `toString()` includes the raw `apiKey` field. If anyone ever logs the config object (e.g. `log.info { "config=$config" }`) or it appears in a stack trace, the key is exposed in plaintext.

Current code carefully logs only specific fields (SystemUserInitializer L24), but this is fragile — future changes could inadvertently log the whole object.

Suggest overriding `toString()` to mask it:

```kotlin
override fun toString() = "SystemUserConfig(githubId=$githubId, name=$name, email=$email, apiKey=***)"
```

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -47,6 +47,28 @@ class ApiKeyModel {
         return GeneratedApiKey(key = rawKey, createdAt = now)
     }
 
+    /**
+     * Upsert a given API key. This is only used by the init function where the user can pass
+     * in an API key for a system user.
+     */
+    fun upsertApiKey(userId: Long, rawKey: String) {
```
fengelniederhammer, line 54: 🟡 risk: `upsertApiKey` accepts any string, including empty. If the `SYSTEM_USER_API_KEY` env var is unset or empty in the deployment config, the system user gets a key whose SHA-256 hash is the well-known `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` (hash of empty string). Anyone who knows this could authenticate as the system user.

Suggest adding a minimum length check:

```kotlin
fun upsertApiKey(userId: Long, rawKey: String) {
    require(rawKey.length >= 32) { "API key must be at least 32 characters" }
    ...
}
```

backend/src/test/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializerTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    private fun uniqueKey() = UUID.randomUUID().toString()
+
+    private fun initializerWith(config: SystemUserConfig?) = SystemUserInitializer(
+        dashboardsConfig = DashboardsConfig(organisms = emptyMap(), systemUser = config),
+        userModel = userModel,
+        apiKeyModel = apiKeyModel,
+    )
+
+    @Test
+    fun `WHEN systemUser config with apiKey THEN key validates and user exists`() {
```
fengelniederhammer: 🔵 nit: Test names use `WHEN...THEN` format but project convention (per `backend/AGENTS.md` and all other controller tests) requires `` `GIVEN ... WHEN ... THEN ...` `` backtick format. E.g. `` `GIVEN systemUser config with apiKey WHEN initializer runs THEN key validates and user exists` ``.

backend/src/test/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializerTest.kt
fengelniederhammer, line 1: ❓ question: Missing test for `SystemUserConfig` with `apiKey = null`. The `config.apiKey?.let` path (SystemUserInitializer L25) — user created but no key upserted — is untested. Consider adding:

```kotlin
@Test
fun `GIVEN systemUser config without apiKey WHEN initializer runs THEN user is created but no key exists`() {
    val githubId = "system-user-no-key-${System.nanoTime()}"
    initializerWith(SystemUserConfig(githubId = githubId, name = "Bot", email = null, apiKey = null)).run(noArgs)
    // verify user exists, verify GET /api-keys returns 404
}
```

fengelniederhammer (commented):
🟡 **Account identity collision via `githubId`**

The `systemUser.githubId` in the YAML config uses a real GitHub user ID. Since `syncUser` is keyed on `githubId`, the system user and any OAuth user who logs in with the same GitHub ID are **the same account**. This means:

- If the configured `githubId` belongs to a real person, the system API key grants full access to that person's collections.
- Every backend restart, `SystemUserInitializer` overwrites the user's `name` and `email` back to the config values. Conversely, every OAuth login overwrites them to the GitHub profile values.
- If someone misconfigures `githubId` to another user's ID, the pre-seeded API key becomes an access key to that user's resources.

This isn't a bug per se (the intent is to use a dedicated bot account), but it should be documented explicitly: the configured `githubId` **must** be a dedicated bot/service account, never a real human user's ID.

---

Other findings are in the individual review comments above (toString key leak, empty key hash, test naming, missing test for null apiKey).

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
```diff
@@ -47,6 +47,28 @@ class ApiKeyModel {
...
+                this.keyHash = hash
+                this.createdAt = now
+                this.lastUsedAt = null
+            }
+        } else if (existing.keyHash != hash) {
+            existing.keyHash = hash
+            existing.createdAt = now
+            existing.lastUsedAt = null
+        }
+    }
```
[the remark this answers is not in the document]
fhennig: We are only running a single instance, so this is fine.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/config/SystemUserInitializer.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun run(args: ApplicationArguments) {
+        val config = dashboardsConfig.systemUser ?: return
+        val user = userModel.syncUser(
+            UserSyncRequest(githubId = config.githubId, name = config.name, email = config.email),
+        )
+        log.info { "System user ready: id=${user.id}, githubId=${config.githubId}, name=${config.name}" }
+        config.apiKey?.let {
+            apiKeyModel.upsertApiKey(userId = user.id, rawKey = it)
+            log.info { "System user API key upserted for userId=${user.id}" }
+        }
```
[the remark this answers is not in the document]
fhennig: I've added a check instead so we cannot have empty keys.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/apikey/ApiKeyModel.kt
fhennig: check added elsewhere.

fengelniederhammer (approved):

PR #1221, opened 2026-05-20
feat(backend): add pagination to GET /collections

fhennig:
## Summary

- Adds `page` (1-based, default 1) and `pageSize` (default 100) query params to `GET /collections`
- Changes the response from `List<Collection>` to `PaginatedResponse<Collection>` — a `{ data, totalCount, page, pageSize }` envelope
- Adds `PaginatedResponse<T>` data class to `api/Collection.kt`
- Updates `CollectionsController`, `CollectionModel`, `CollectionsClient` (test helper), and all existing tests to use `.data` on the response
- Adds 4 new pagination-specific tests

## Current state — does not compile

The implementation is blocked on the correct Exposed 1.3.0 API for **offset/limit pagination** on a `Query`. Specifically:

1. **`limit(n, offset)`** — In Exposed 1.x, `SizedIterable.limit()` and `Query.limit()` only accept a single `count: Int` argument. The old `limit(n, offset)` two-argument form no longer exists. A separate `.offset(n)` method may exist on `Query`/`AbstractQuery`, but hasn't been verified.

2. **`selectAll()` / `andWhere {}`** — The raw query DSL entry point and the method to add WHERE clauses incrementally need the correct import paths for Exposed 1.3.0. `selectAll` resolved via `import org.jetbrains.exposed.v1.jdbc.selectAll`, but `andWhere` is still unresolved.

### What needs to be figured out

The right Exposed 1.x pattern for: "build a query with optional WHERE conditions, get the total count, then fetch one page with LIMIT + OFFSET". Likely something like:

```kotlin
// Option A — build condition upfront, use where { }, chain offset()
var cond: Op<Boolean> = Op.TRUE
if (userId != null) cond = cond and (CollectionTable.ownedBy eq userId)
if (organism != null) cond = cond and (CollectionTable.organism eq organism)
val query = CollectionTable.selectAll().where { cond }
val totalCount = query.count()
val paged = CollectionEntity.wrapRows(query.orderBy(...).limit(pageSize).offset(offset))

// Option B — go back to EntityClass.find { } and figure out offset
CollectionEntity.find { cond }.orderBy(...).limit(pageSize).offset(offset)
```

## Test plan
- [ ] Resolve compile error (correct Exposed 1.x pagination API)
- [ ] Run `./gradlew test --tests "*.CollectionsGetTest"` — all existing + 4 new tests pass
- [ ] Manually verify `GET /collections?page=2&pageSize=10` returns correct slice and `totalCount`

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
@@ -84,3 +84,10 @@ data class CollectionUpdate(
     val description: String? = null,
     val variants: List<VariantUpdate>? = null,
 )
+
+data class PaginatedResponse<T>(
+    val data: List<T>,
+    val totalCount: Long,
+    val page: Int,
+    val pageSize: Int,
+)
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsController.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsController.kt
@@ -5,6 +5,7 @@ import io.swagger.v3.oas.annotations.Parameter
 import org.genspectrum.dashboardsbackend.api.Collection
 import org.genspectrum.dashboardsbackend.api.CollectionRequest
 import org.genspectrum.dashboardsbackend.api.CollectionUpdate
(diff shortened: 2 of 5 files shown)


fhennig:
We won't need this any time soon.

Instead, we want to make it possible to omit variants from the collections, which saves a lot of space. And also make it possible to filter better the collections.

PR #1229, opened 2026-05-27
feat(website): predefined COVID variant signatures from collections (W-ASAP)

### LINKED ISSUE #1199: wastewater: allow using Cornelius' definitions of variants
Cornelius maintains a list of variants and explicit mutations that with the variant in this repo: https://github.com/corneliusroemer/pango-sequences/

raw file: https://raw.githubusercontent.com/corneliusroemer/pango-sequences/refs/heads/main/data/pango-consensus-sequences_summary.json

Currently, we 'calculate' the mutation signature of a variant:

Image

For various reasons, this is not ideal.

We want users to be able to instead pick the curated lists of mutations to look at. We would still allow users the 'old' way of statistically computing the variant signature.

### Authentication
This depends on #1200 (API key auth). A dedicated bot user will be created; its API key stored as a GitHub Secret and passed as `API_KEY` env var to the container.

### Approach: containerized ingest scripts

Ingest scripts live in this repository (under `ingest/`) as one or more containers. Starting point is the existing `example-data/seed-hello-world.mjs`, which already calls the collections API — it just needs to swap session-cookie auth for `Authorization: Bearer $API_KEY`.

A single container with subcommands (e.g. `node ingest.mjs pango-sequences`) is the default, but separate containers per source are also fine if different runtimes are needed (e.g. Python for one source, JS for another).

**Overwrite strategy:** on each run, the script overwrites the target collection in full. No diffing needed.

### Pango-sequences script
Reads [Cornelius' variant definitions](https://github.com/corneliusroemer/pango-sequences/) and upserts a collection via the API.

### Running the container

**Staging (on DB reset):** added as a short-lived service in `docker-compose` (e.g. `restart: no`), runs automatically when staging is reset.

**Production (regular updates):** GitHub Actions runs the container on a schedule, passing `API_KEY` from GitHub Secrets and the prod API URL.

fhennig:
resolves #1199 

Adds support for predefined variant signatures in the W-ASAP Variant Explorer mode, as an alternative to computing signatures from clinical sequences. For COVID, these come from a configured collections source (filtered by user ID, organism, and a tag in the description).

### Summary

- Adds a `predefinedVariantsSource` config to `WasapPageConfig` specifying where to fetch predefined variant collections from (userId, tag, label); the organism is taken from the new `internalName` field on the base config
- Pre-fetches matching collections via `getCollectionSummaries` and exposes them in a **downshift autocomplete combobox** in the filter panel
- Extends `WasapVariantFilter` with a `signatureType` field (`'computed'` | `'predefined'`) and predefined-specific fields (`collectionId`, `newMutationsOnly`); serialised to/from URL
- `signatureType` defaults to `'computed'` and is omitted from the URL for backwards compatibility with existing bookmarks
- Refactors the variant filter type from a discriminated union to a flat type (consistent with `WasapUntrackedFilter`), removing the need to pass `computedDefaults` through the component tree
- Extracts `SignatureType` to `wasapPageConfig.ts` alongside all other WASAP types
- Adds `internalName` to `WasapPageConfigBase` so the organism key is accessible from the config directly
- Updates the Variant Explorer info box to describe both predefined and computed signature modes

### Screenshot

image

### Testing

- **`WasapPageStateHandler.spec.ts`** — URL round-trip tests for the predefined `signatureType`: serialising a `WasapVariantFilter` with `signatureType: 'predefined'`, `collectionId`, and `newMutationsOnly` into URL params and parsing them back produces the original state.
- **`CollectionCombobox.browser.spec.tsx`** (new file) — Browser tests for the autocomplete combobox used to pick a predefined collection: selecting a collection from the dropdown, clearing the selection via the clear button, and blur behaviour (committing a valid value on blur, rejecting an invalid one with an error state).
- **`VariantExplorerFilter.browser.spec.tsx`** — Browser tests for the predefined path in the filter panel: switching the "Variant definition source" dropdown from computed → predefined, selecting a collection, and toggling the "Mutation not in parent" checkbox.

## PR Checklist

### DIFF
--- a/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
@@ -112,9 +112,9 @@ export function WasapPageStateSelector({
                     "This predefined variants query was called despite it being disabled. This should not happen.",
                 );
             }
-            const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
+            const { collectionsUserId, collectionsOrganism, collectionsTag } = config.predefinedVariantsSource;
             return getBackendServiceForClientside()
-                .getCollections({ userId: collectionsUserId, organism: config.name })
+                .getCollectionSummaries({ userId: collectionsUserId, organism: collectionsOrganism })
                 .then((collections) => collections.filter((c) => c.description?.includes(collectionsTag) ?? false));
         },
     });
--- a/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
@@ -1,6 +1,8 @@
 import { useQuery } from '@tanstack/react-query';
 import { type Dispatch, type SetStateAction, useState } from 'react';
 
+import { getBackendServiceForClientside } from '../../../backendApi/backendService';
+
 import { ApplyFilterButton } from '../ApplyFilterButton';
 import { DynamicDateFilter } from '../DynamicDateFilter';
 import { SelectorHeadline } from '../SelectorHeadline';
@@ -101,6 +103,22 @@ export function WasapPageStateSelector({
         },
     });
 
+    const predefinedVariantsQueryResult = useQuery({
+        enabled: config.variantAnalysisModeEnabled && config.predefinedVariantsSource !== undefined,
+        queryKey: ['predefinedVariants', config.variantAnalysisModeEnabled && config.predefinedVariantsSource],
+        queryFn: () => {
+            if (!config.variantAnalysisModeEnabled || config.predefinedVariantsSource === undefined) {
+                throw Error(
+                    "This predefined variants query was called despite it being disabled. This should not happen.",
+                );
+            }
+            const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
+            return getBackendServiceForClientside()
+                .getCollections({ userId: collectionsUserId, organism: config.name })
+                .then((collections) => collections.filter((c) => c.description?.includes(collectionsTag) ?? false));
+        },
+    });
+
     return (
         <div className='flex flex-col gap-4'>
             <SelectorHeadline>Filter dataset</SelectorHeadline>
@@ -181,6 +199,11 @@ export function WasapPageStateSelector({
                                     setPageState={setVariantFilter}
                                     clinicalSequenceLapisBaseUrl={config.clinicalLapis.lapisBaseUrl}
                                     clinicalSequenceLapisLineageField={config.clinicalLapis.lineageField}
+                                    predefinedVariantsQueryResult={
+                                        config.predefinedVariantsSource !== undefined
+                                            ? predefinedVariantsQueryResult
+                                            : undefined
+                                    }
                                 />
                             );
                         case 'resistance':
--- a/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
@@ -111,9 +111,9 @@ export function WasapPageStateSelector({
                     'This predefined variants query was called despite it being disabled. This should not happen.',
                 );
             }
-            const { collectionsUserId, collectionsOrganism, collectionsTag } = config.predefinedVariantsSource;
+            const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
             return getBackendServiceForClientside()
-                .getCollectionSummaries({ userId: collectionsUserId, organism: collectionsOrganism })
+                .getCollectionSummaries({ userId: collectionsUserId, organism: config.internalName })
                 .then((collections) => collections.filter((c) => c.description?.includes(collectionsTag) ?? false));
         },
     });
--- a/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
@@ -2,7 +2,6 @@ import { useQuery } from '@tanstack/react-query';
 import { type Dispatch, type SetStateAction, useState } from 'react';
 
 import { getBackendServiceForClientside } from '../../../backendApi/backendService';
-
 import { ApplyFilterButton } from '../ApplyFilterButton';
 import { DynamicDateFilter } from '../DynamicDateFilter';
 import { SelectorHeadline } from '../SelectorHeadline';
@@ -109,7 +108,7 @@ export function WasapPageStateSelector({
         queryFn: () => {
             if (!config.variantAnalysisModeEnabled || config.predefinedVariantsSource === undefined) {
                 throw Error(
-                    "This predefined variants query was called despite it being disabled. This should not happen.",
+                    'This predefined variants query was called despite it being disabled. This should not happen.',
                 );
             }
             const { collectionsUserId, collectionsOrganism, collectionsTag } = config.predefinedVariantsSource;
@@ -204,6 +203,7 @@ export function WasapPageStateSelector({
                                             ? predefinedVariantsQueryResult
                                             : undefined
                                     }
+                                    predefinedVariantsLabel={config.predefinedVariantsSource?.variantSourceLabel}
                                 />
                             );
                         case 'resistance':
--- a/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
+++ b/website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
@@ -105,16 +105,18 @@ export function WasapPageStateSelector({
     const predefinedVariantsQueryResult = useQuery({
         enabled: config.variantAnalysisModeEnabled && config.predefinedVariantsSource !== undefined,
         queryKey: ['predefinedVariants', config.variantAnalysisModeEnabled && config.predefinedVariantsSource],
-        queryFn: () => {
+        queryFn: async () => {
             if (!config.variantAnalysisModeEnabled || config.predefinedVariantsSource === undefined) {
                 throw Error(
                     'This predefined variants query was called despite it being disabled. This should not happen.',
                 );
             }
             const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
-            return getBackendServiceForClientside()
-                .getCollectionSummaries({ userId: collectionsUserId, organism: config.internalName })
-                .then((collections) => collections.filter((c) => c.description?.includes(collectionsTag) ?? false));
+            const collections = await getBackendServiceForClientside().getCollectionSummaries({
+                userId: collectionsUserId,
+                organism: config.internalName,
+            });
+            return collections.filter((c) => c.description?.includes(collectionsTag) ?? false);
         },
     });
 
--- a/website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
+++ b/website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
@@ -1,23 +1,21 @@
 import { type UseQueryResult } from '@tanstack/react-query';
-import { useState } from 'react';
 
-import { type CollectionSummary } from '../../../../types/Collection';
-import { CollectionCombobox } from '../utils/CollectionCombobox';
 import { Inset } from '../../../../styles/Inset';
+import { type CollectionSummary } from '../../../../types/Collection';
 import { GsLineageFilter } from '../../../genspectrum/GsLineageFilter';
 import {
     VARIANT_TIME_FRAME,
     variantTimeFrameLabel,
+    type SignatureType,
     type VariantTimeFrame,
     type WasapVariantFilter,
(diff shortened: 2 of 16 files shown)


fhennig:
feedback re labels:

image

website/src/components/views/wasap/useWasapPageData.ts
```diff
@@ -68,6 +69,21 @@ function fetchManualModeData(config: WasapPageConfig, analysis: WasapManualFilte
...
+    if (!config.variantAnalysisModeEnabled) {
+        throw Error("Cannot fetch data, 'variant' mode is not enabled.");
+    }
+    switch (analysis.signatureType) {
+        case 'computed':
+            return fetchVariantComputedModeData(config, analysis);
+        case 'predefined':
+            return fetchVariantPredefinedModeData(analysis);
+    }
+}
```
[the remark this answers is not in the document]
fhennig: Shouldn't happen, we're guarded against that.

website/src/components/pageStateSelectors/wasap/utils/CollectionCombobox.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    <button
+                        aria-label='clear selection'
+                        className='px-2'
+                        type='button'
+                        tabIndex={-1}
+                        onClick={() => {
+                            reset();
+                            onChange(null);
+                        }}
+                    >
```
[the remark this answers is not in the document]
fhennig: we want it that way, it's in line with the other components. So you can cycle through the filters.

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -52,12 +53,15 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
...
                 if (!this.config.variantAnalysisModeEnabled) {
                     throw Error("The 'variant' analysis mode is not enabled.");
                 }
                 analysis = {
                     mode,
+                    signatureType:
+                        (texts.signatureType as SignatureType | undefined) ??
+                        this.config.filterDefaults.variant.signatureType,
                     sequenceType: providedSequenceType ?? this.config.filterDefaults.variant.sequenceType,
                     variant: texts.variant ?? this.config.filterDefaults.variant.variant,
```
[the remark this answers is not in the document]
fhennig: no.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
```diff
@@ -21,76 +26,151 @@ interface VariantExplorerFilterProps {
...
-                />
-                <LabeledField label='Time frame'>
+            {predefinedVariantsQueryResult !== undefined && (
+                <LabeledField label='Variant definition source'>
                     <select
                         className='select select-bordered'
-                        value={pageState.timeFrame}
-                        onChange={(e) => setPageState({ ...pageState, timeFrame: e.target.value as VariantTimeFrame })}
+                        value={pageState.signatureType}
+                        onChange={(e) => handleSignatureTypeChange(e.target.value as SignatureType)}
```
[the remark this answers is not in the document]
fhennig: LabeledField is a label.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
```diff
@@ -21,76 +26,151 @@ interface VariantExplorerFilterProps {
...
+    const selectedCollection = collections.find((c) => c.id === pageState.collectionId) ?? null;
+
+    return (
+        <Inset className='mt-4 p-2'>
+            <LabeledField label='Variant'>
+                <CollectionCombobox
+                    collections={collections}
+                    value={selectedCollection}
+                    onChange={(c) => setPageState({ ...pageState, collectionId: c?.id })}
+                />
```
fengelniederhammer, line 162: https://github.com/user-attachments/assets/64f2f78d-d359-43b3-8348-855233aa1582

Filtering for the first letter is quite slow

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
```diff
@@ -21,76 +26,151 @@ interface VariantExplorerFilterProps {
...
+                <input
+                    className='accent-primary'
+                    type='checkbox'
+                    id='newMutationsOnly'
+                    checked={pageState.newMutationsOnly ?? false}
+                    onChange={(e) => setPageState({ ...pageState, newMutationsOnly: e.target.checked })}
+                />
+                <label htmlFor='newMutationsOnly' className='pl-2'>
+                    Mutation not in parent
+                </label>
```
fengelniederhammer: Should we maybe add a tooltip or something that explain what it does?

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -52,12 +53,15 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
...
-            case 'variant':
+            case 'variant': {
                 if (!this.config.variantAnalysisModeEnabled) {
                     throw Error("The 'variant' analysis mode is not enabled.");
                 }
                 analysis = {
                     mode,
+                    signatureType:
+                        (texts.signatureType as SignatureType | undefined) ??
+                        this.config.filterDefaults.variant.signatureType,
```
fengelniederhammer, line 64: Do we need to validate that this has only one of the expected values?

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -135,11 +144,23 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
...
+                    if (analysis.newMutationsOnly) {
+                        setSearchFromString(search, 'newMutationsOnly', 'true');
+                    }
+                } else {
+                    setSearchFromString(search, 'variant', analysis.variant);
+                    setSearchFromString(search, 'minProportion', String(analysis.minProportion));
+                    setSearchFromString(search, 'minCount', String(analysis.minCount));
+                    setSearchFromString(search, 'minJaccard', String(analysis.minJaccard));
+                    setSearchFromString(search, 'timeFrame', analysis.timeFrame);
+                }
```
fengelniederhammer, line 163: Is this intentionally not setting the `signatureType`? (Also works without it apparently, but it's a bit surprising)

website/src/components/views/wasap/WasapPage.tsx
fengelniederhammer, line 1: Applying filters without a collection yields this quite generic error message:

Image

Might be worth improving.

website/src/components/pageStateSelectors/wasap/WasapPageStateSelector.tsx
```diff
@@ -101,6 +102,22 @@ export function WasapPageStateSelector({
...
+            if (!config.variantAnalysisModeEnabled || config.predefinedVariantsSource === undefined) {
+                throw Error(
+                    'This predefined variants query was called despite it being disabled. This should not happen.',
+                );
+            }
+            const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
+            return getBackendServiceForClientside()
+                .getCollectionSummaries({ userId: collectionsUserId, organism: config.internalName })
+                .then((collections) => collections.filter((c) => c.description?.includes(collectionsTag) ?? false));
+        },
```
fengelniederhammer: My IDE suggests
```suggestion
        queryFn: async () => {
            if (!config.variantAnalysisModeEnabled || config.predefinedVariantsSource === undefined) {
                throw Error(
                    'This predefined variants query was called despite it being disabled. This should not happen.',
                );
            }
            const { collectionsUserId, collectionsTag } = config.predefinedVariantsSource;
            const collections = await getBackendServiceForClientside().getCollectionSummaries({
                userId: collectionsUserId,
                organism: config.internalName,
            });
            return collections.filter((c) => c.description?.includes(collectionsTag) ?? false);
        },
```

website/src/components/views/wasap/useWasapPageData.ts
```diff
@@ -97,6 +113,36 @@ async function fetchVariantModeData(
     };
 }
 
+async function fetchVariantPredefinedModeData(analysis: WasapVariantFilter): Promise<WasapMutationsData> {
+    if (!analysis.collectionId) {
```
fengelniederhammer: AI: 🟡 risk: `!analysis.collectionId` is falsy for `0`. If collection IDs ever start at 0, this silently throws. Use `analysis.collectionId == null` instead.

fengelniederhammer: (This is exactly why I always try to avoid those implicit casts)

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fhennig: Good find! I only tested with 10 collections so I didn't realize that this was an issue.

Yeah I think the issue is that we're displaying 5000 entries in the dropdown. Not sure what the best way forward is.

Should we add a virtualization library just for this component?

We could also cap the amount of things we display, which would be quite simple to implement. No one can realistically scroll through 1000s of entries anyways; I'd assume that 99% of users type something in that they already know. I think I'd cap, and maybe we can add a little note at the end of the list saying "not all entries shown; type more to filter selection" or something along those lines.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fhennig: done.

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
fhennig: We're not doing any runtime validation in the handler in general, so I think we don't need to do it here either.

website/src/components/views/wasap/WasapPage.tsx
fhennig: Ah yes, damn. I think it used to be the case that if you would apply the filters in this state, once we parse from the URL, a default value would be set again. But I think since it's all one component now, maybe this doesn't happen anymore? 

I also see weird behaviour on the 'computed' mode, when no variant is selected. We used to prevent the "no variant selected" case by just picking a default, but I think that mightnot work anymore, actually. The state is pushed directly and we don't go through the handler anymore.

We need a different way to prevent the user from just not having anything selected.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fhennig: ... I tried looking at our GsLineageFilter, because it doesn't have the same problem. It renders a few thousand elements and searches them without issues. But looking at the implementation, I can't really find any impactful differences. The only thing I can think of now, is that we're using Preact and the web components, instead of React. Could that explain it?

I still think that probably capping the amount of elements in the dropdown is the way to solve this here.

website/src/components/views/wasap/WasapPage.tsx
fhennig: An example of how we could implement this in a simple way: https://github.com/GenSpectrum/dashboards/pull/1253

But I think maybe it would be best to do this in the components already.

For this particular `CollectionCombox` I think we can actually implement it so that a value is always selected, that undefined is just not selectable. No clear button.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fengelniederhammer: There is also a hook `useId()` that we could use instead of hardcoding the id.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fengelniederhammer: I'm also not aware that we have a specific performance optimizatin in the gs-lineage-filter... I also don't have a good idea right now how to solve this.

website/src/components/views/wasap/WasapPage.tsx
fengelniederhammer: > For this particular `CollectionCombox` I think we can actually implement it so that a value is always selected, that undefined is just not selectable. No clear button.

Didn't fully think it through, but sounds promising to me :+1:

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fhennig: Update: We found that this is an issue only in Firefox; in Chrome and Safari, the box filters instantly. From profiling it also seemed to render fast in React (~30ms) but then subsequently be slow in the Firefox.

We decided that we will now just accept the slight lag in Firefox, until someone complains.

website/src/components/views/wasap/WasapPage.tsx
fhennig: i've solved it like this now:

image

fengelniederhammer (approved):
LGTM (not sure why the diff shows 31 commits now, looks like something went wrong during the rebase)

PR #1234, opened 2026-05-30
feat: add excludeSystemCollections param to GET /collections

### LINKED ISSUE #1231: Deal with the many generated collections
Now that we have almost 5k generated collections, the list in the UI is very long.

We either need to add at least frontend pagination, or we thought about just hiding the system-user collections in the UI - that might actually be the easiest thing to do.

fhennig:
resolves #1231 

## Summary

- Adds `excludeSystemCollections` query parameter to `GET /collections` (default `false`)
- The collections overview table in the frontend always passes `excludeSystemCollections=true`

## Screenshot

**No collections in the list, but individual collections can still be accessed**

https://github.com/user-attachments/assets/4c909dd3-2755-47c4-ac7b-55584b1fe222

## Test plan

- [x] `GET /collections` without the param returns all collections including system-owned ones (no regression)
- [x] `GET /collections?excludeSystemCollections=true` excludes collections owned by the system user
- [x] If no system user is configured, the param has no effect
- [x] Collections overview in the UI no longer shows system-owned collections

### DIFF
--- a/backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsGetTest.kt
+++ b/backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsGetTest.kt
@@ -301,6 +301,16 @@ class CollectionsGetTest(
     }
 
     @Test
+    fun `WHEN no system user configured THEN excludeSystemCollections=true returns all collections`() {
+        val userId = usersClient.createUser()
+        val collection = collectionsClient.postCollection(dummyCollectionRequest.copy(name = "No System User"), userId)
+
+        val collections = collectionsClient.getCollections(userId = userId, excludeSystemCollections = true)
+
+        assertThat(collections, hasItem(collection))
+    }
+
+    @Test
     fun `WHEN getting collections with includeVariants=true THEN variantCount and variants are both present`() {
         val userId = usersClient.createUser()
         collectionsClient.postCollection(dummyCollectionRequest, userId)
--- a/backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsGetTest.kt
+++ b/backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsGetTest.kt
@@ -305,7 +305,11 @@ class CollectionsGetTest(
         val userId = usersClient.createUser()
         val collection = collectionsClient.postCollection(dummyCollectionRequest.copy(name = "No System User"), userId)
 
-        val collections = collectionsClient.getCollections(userId = userId, excludeSystemCollections = true)
+        val collections = collectionsClient.getCollections(
+            userId = userId,
+            includeVariants = true,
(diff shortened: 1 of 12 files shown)


backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsGetTest.kt
```diff
@@ -300,6 +300,20 @@ class CollectionsGetTest(
             .andExpect(jsonPath("$[0].variants").doesNotExist())
     }
 
+    @Test
+    fun `WHEN no system user configured THEN excludeSystemCollections=true returns all collections`() {
```
fengelniederhammer:
```suggestion
    fun `GIVEN no system user configured WHEN getting with excludeSystemCollectionsTHEN returns all collections`() {
```

website/src/backendApi/backendService.ts
```diff
@@ -177,8 +177,13 @@ export class BackendService extends ApiService {
...
 
-    public async getCollectionSummaries({ organism }: { organism?: string } = {}) {
-        const requestParams = organism !== undefined ? { organism } : undefined;
+    public async getCollectionSummaries({
+        organism,
+        excludeSystemCollections = false,
+    }: { organism?: string; excludeSystemCollections?: boolean } = {}) {
+        const requestParams: Record<string, string> = {};
+        if (organism !== undefined) requestParams.organism = organism;
+        if (excludeSystemCollections) requestParams.excludeSystemCollections = 'true';
```
fengelniederhammer:
```suggestion
    public async getCollectionSummaries(requestParams: { organism?: string; excludeSystemCollections?: boolean } = {}) {
```
I think this also works?

website/routeMocker.ts
```diff
@@ -34,7 +34,16 @@ export class AstroApiRouteMocker {
...
             http.get(
                 `${ASTRO_SERVER_URL}/collections`,
-                resolver([{ statusCode, response, requestParam: organism !== undefined ? { organism } : undefined }]),
+                resolver([
+                    {
+                        statusCode,
+                        response,
+                        requestParam: {
+                            ...(organism !== undefined ? { organism } : {}),
+                            excludeSystemCollections: 'true',
```
fengelniederhammer: AI: 🔵 nit: `AstroApiRouteMocker.mockGetCollectionSummaries` now hard-codes `excludeSystemCollections: 'true'` in the expected query params, but `BackendRouteMocker.mockGetCollectionSummaries` (L228) wasn't updated — it still doesn't expect the param. This asymmetry is inconsistent. Consider adding an optional `excludeSystemCollections` argument to both mockers (or at least aligning them).

fengelniederhammer: Aha good catch. I think this function is a misplaced duplicate anyway and can probably be deleted?

fhennig: both changed now, I believe.

fengelniederhammer (approved):

PR #1236, opened 2026-06-01
docs: add ARC42 section 5 - Block view

fhennig:
**See [here](https://github.com/GenSpectrum/dashboards/blob/docs%2Farc-block-view/docs/arc42/05-building-block-view.md)**

### Summary

Add the section 5 as it was missing - obviously not entirely complete as you can always go deeper.


docs/arc42/05-building-block-view.md
```diff
@@ -0,0 +1,24 @@
...
+## Level 0: System Context
+
+This diagram shows GenSpectrum in the context of its neighboring systems.
+
+![overview](../plantuml/05-overview.svg)
+
+* Users interact with GenSpectrum via the website.
+* Genomic sequence data is queried from Pathoplexus.
+* Raw sequence data is pulled from NCBI.
+* GitHub is used for user authentication via OAuth.
```
fengelniederhammer, line 12: This has some significant overlap with 03-context, maybe can deduplicate some of it?

Also, IMO, we shouldn't document "GenSpectrum" (as in our whole server setup landscape) here, but the artifacts in this repo.

docs/plantuml/05-overview.puml
```diff
@@ -0,0 +1,16 @@
...
+actor User as user
+
+component "GenSpectrum" as gs
+component "NCBI" as ncbi
+component "Pathoplexus" as pathoplexus
+component "GitHub" as github
+
+user -r-> gs
+gs -d-> ncbi : "Pull data"
+gs -d-> pathoplexus : "Query genomic data"
```
fengelniederhammer, line 13: It's not only Pathoplexus that we depend on, it's all sorts of LAPISes.

docs/plantuml/05-overview.puml
```diff
@@ -0,0 +1,16 @@
...
+
+actor User as user
+
+component "GenSpectrum" as gs
+component "NCBI" as ncbi
+component "Pathoplexus" as pathoplexus
+component "GitHub" as github
+
+user -r-> gs
+gs -d-> ncbi : "Pull data"
```
fengelniederhammer, line 12: We don't really have a direct dependency on NCBI in these artifacts, do we?

docs/plantuml/05-level1.puml
```diff
@@ -0,0 +1,33 @@
...
+component "GitHub" <<external>> as github
+component "Pathoplexus" <<external>> as pathoplexus
+cloud "Various external mutation information" as vemi
+
+node "GenSpectrum" {
+    component Website as website
+    component Backend as backend
+    component "Collection Seeder" as seeder
+    database Database as db
+    component Loculus as loculus
```
fengelniederhammer, line 19: IMO Loculus shouldn't be in here, it's external to this repo.

fengelniederhammer (changes requested):
The servers repo contains higher-level documentation (how different deployment units interact). Some of those things should be documented there (and probably already are).

PR #1254, opened 2026-06-08
fix(github-actions): restore closing quote on body string in rebaseProd workflow

fhennig:
## Summary

- `0a186a8` removed the closing `"` from the `body` variable when updating the PR message wording
- This caused the shell to treat the `gh pr create` line as part of the string, erroring with `unexpected EOF while looking for matching '"'`
- Restores the closing quote and moves `gh pr create` back outside the string


fengelniederhammer (approved):
Thanks :)

PR #1262, opened 2026-06-09
fix(website): collection page URLs use pathFragment instead of organism enum value

### LINKED ISSUE #1258: Collection page URLs use organism enum value instead of pathFragment
## Problem

Collection page URLs are built using the `Organism` enum value directly, rather than the `pathFragment` used everywhere else in the app.

In `src/types/pages.ts`:
```ts
collectionsForOrganism: (organism: Organism) => `/collections/${organism}`,
viewCollection: (organism: Organism, id: string) => `/collections/${organism}/${id}`,
editCollection: (organism: Organism, id: string) => `/collections/${organism}/${id}/edit`,
createCollection: (organism: Organism) => `/collections/${organism}/create`,
```

This produces URLs like `/collections/westNile` and `/collections/rsvA`, while the rest of the site uses the `pathFragment` from `organismConfig` (e.g. `west-nile`, `rsv-a`).

Affected organisms where enum key ≠ pathFragment: `westNile`, `rsvA`, `rsvB`, `ebolaSudan`, `ebolaZaire`, `influenzaA`, `influenzaB`.

## Fix

1. Rename the Astro route folder `collections/[organism]` → `collections/[pathFragment]`
2. Update `pages.ts` to use `organismConfig[organism].pathFragment` in all four collection URL builders
3. In each of the four Astro collection pages, inline a lookup from pathFragment → Organism:
   ```astro
   const { pathFragment } = Astro.params;
   const organism = allOrganisms.find(o => organismConfig[o].pathFragment === pathFragment);
   if (!organism) return Astro.redirect('/404');
   ```

Note: backend API calls already correctly use the `Organism` enum value (matching the YAML config keys), so no backend changes are needed.

fhennig:
Closes #1258

## Summary

- Renames the Astro route directory `collections/[organism]` → `collections/[pathFragment]` so the URL segment is the kebab-case path fragment (e.g. `west-nile`, `rsv-a`) consistent with the rest of the site
- Updates the four URL builders in `pages.ts` to use `organismConfig[organism].pathFragment` instead of the raw enum key
- Updates all four collection Astro pages to resolve `pathFragment` → `Organism` via `allOrganisms.find()`

## Test plan

- [x] Visit `/collections/west-nile`, `/collections/rsv-a`, `/collections/rsv-b`, `/collections/ebola-sudan`, `/collections/ebola-zaire`, `/collections/influenza-a`, `/collections/influenza-b` and verify they load correctly
- [x] Verify the old URLs (e.g. `/collections/westNile`) now 404
- [x] Verify collection detail, edit, and create pages work for the affected organisms
- [x] Verify organisms where enum key = pathFragment (e.g. `covid`, `mpox`) are unaffected

### DIFF
--- a/website/src/pages/collections/[pathFragment]/[id]/edit.astro
+++ b/website/src/pages/collections/[pathFragment]/[id]/edit.astro
@@ -6,16 +6,16 @@ import { getDashboardsConfig, getBackendHost } from '../../../../config';
 import { defaultBreadcrumbs } from '../../../../layouts/Breadcrumbs';
 import ContaineredPageLayout from '../../../../layouts/ContaineredPage/ContaineredPageLayout.astro';
 import { getInstanceLogger } from '../../../../logger.ts';
-import { organismConfig, organismSchema } from '../../../../types/Organism';
+import { allOrganisms, organismConfig } from '../../../../types/Organism';
 import { Page } from '../../../../types/pages';
 import { getErrorLogMessage } from '../../../../util/getErrorLogMessage.ts';
 
 const logger = getInstanceLogger('CollectionEditPage');
 
-const { organism, id } = Astro.params;
+const { pathFragment, id } = Astro.params;
 
-const parsedOrganism = organismSchema.safeParse(organism);
-if (!parsedOrganism.success) {
+const organism = allOrganisms.find((o) => organismConfig[o].pathFragment === pathFragment);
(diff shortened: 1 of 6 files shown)


fengelniederhammer (commented):
Looks good, but let's fix the type error ;)

fengelniederhammer (approved):

PR #1274, opened 2026-06-11
refactor(website): use per-mutation annotations for wastewater resistance mutations

### LINKED ISSUE #1223: Annotations - allow descriptions for individual mutations
Our current definition of an annotation is a title and description, and list of mutations.

For resistance mutations, we want to be able to give information for a particular mutation, somethink like ORF1a:T2343C becomes 3CLpro:T31C (2343 on the gene is 31 in the protein(?)). We currently do this, by having one annotation for each mutation:

Image

With this however we lose the filter functionality by group of mutation:

Image

It is supposed to look like this, the screenshot is from H1N1:

Image

Ideally we could somehow have both. Allow us to filter for a group of mutations that belong together semantically (i.e. all mutations on the Spike protein) but also allow us to have individual information for specific mutations inside the group, so when you click on the little superscript letter, you get information on which particular amino acid in the Spike protein is this.

fhennig:
resolves #1223 

## Summary

- Replaces the two-mode annotation system (`perVariant` / `perCollection`) with a single unified approach, taking advantage of per-mutation name support added in dashboard-components 1.18
- All resistance mutation collections now produce one `MutationAnnotation` per collection, with each individual mutation carrying its variant name via the new `{ mutation, name }` object form
- Removes the `annotationMode` field from `ResistanceMutationCollectionConfig` and the `annotationMode` constant entirely
- Simplifies the test suite from four tests across two `describe` blocks down to two focused tests

## Test plan

- [x] `npm run check-types` passes
- [x] `resistanceData.spec.ts` passes (2 tests)
- [x] `WasapPageStateHandler.spec.ts` passes (26 tests)
- [x] Verified locally against COVID, RSV-A wastewater dashboards

### DIFF
--- a/website/src/components/views/wasap/resistanceData.ts
+++ b/website/src/components/views/wasap/resistanceData.ts
@@ -1,6 +1,6 @@
 import type { MutationAnnotations } from '@genspectrum/dashboard-components/util';
 
-import { annotationMode, type ResistanceMutationCollectionConfig, type WasapPageConfig } from './wasapPageConfig';
+import type { ResistanceMutationCollectionConfig, WasapPageConfig } from './wasapPageConfig';
 import type { BackendService } from '../../../backendApi/backendService';
 import type { Collection } from '../../../types/Collection';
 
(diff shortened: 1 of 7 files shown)


website/src/components/views/wasap/resistanceData.ts
```diff
@@ -52,25 +52,17 @@ export function buildResistanceData(
...
-                description: setConfig.description,
-                aminoAcidMutations: allMutations,
-            });
-        }
+        mutationAnnotations.push({
+            name: setConfig.name,
+            symbol: setConfig.annotationSymbol,
+            description: setConfig.description,
+            aminoAcidMutations: filterVariants.flatMap((variant) =>
+                (variant.filterObject.aminoAcidMutations ?? []).map((mutation) => ({
```
fengelniederhammer, line 60: Can't we also use `allMutations` here?

```suggestion
            aminoAcidMutations:
                allMutations.map((mutation) => ({
```
or something?

fengelniederhammer (commented):
The code looks good, but filtering for any of the mutation sets doesn't return data:

Image

fengelniederhammer (approved):

PR #1277, opened 2026-06-11
fix(collection-seeding): seed placeholder collections to fix CI collection IDs

fhennig:
On fresh DBs, the IDs for the covid resistance mutations are not right, they'd be 1, 2, 3, but on staging and prod they are 4, 5, 6. This causes issues when running E2E tests in CI:

image

We can't change the staging IDs now, and we don't want to introduce yet another switch for different IDs when deploying locally to a fresh, empty DB. My idea is thus, to simply cause the same offset as we have on staging, by inserting placeholder collections.

It's a bit stupid, but it works.

Alternatively, I guess we _could_ wipe the staging DB once, and adjust the IDs in the config accordingly. Maybe that's the better option, actually.

## Summary

- COVID resistance mutation collections are hardcoded as IDs 4, 5, 6 in `wastewaterConfig.ts`
- On a fresh DB (e.g. CI), the seeder assigned them IDs 1, 2, 3 instead, causing the wrong collections to be referenced
- Adds a `PlaceholderSource` that seeds 3 dummy collections first, reserving IDs 1–3 so COVID resistance mutations land on 4, 5, 6 as expected

## Test plan

- [ ] Run collection seeder against a fresh DB and verify COVID resistance mutation collections receive IDs 4, 5, 6
- [ ] Verify CI wastewater dashboard shows correct resistance mutations

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -79,14 +79,16 @@ jobs:
           docker compose -f ../docker-compose.yml logs --no-color "$service" > "/tmp/container-logs/${service}.log" || true
         done
 
-    - uses: actions/upload-artifact@v7
+    - name: Upload container logs
+      uses: actions/upload-artifact@v7
       if: ${{ always() && !cancelled() }}
       with:
         name: container-logs
         path: /tmp/container-logs/
         retention-days: 7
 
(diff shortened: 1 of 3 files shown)


fhennig:
no this is silly, we're not doing this.

PR #1286, opened 2026-06-24
feat(website): show Jaccard index for predefined Nextclade variant mutations

### LINKED ISSUE #240: Provide a Jaccard Index for mutations from Nextclade tree
#234 adds the Nextclade tree as a source of variant definitions, which provide a list of mutations.

We need an indicator of uniqueness of mutations in the Variant Explorer based on the Nextclade tree.

The obvious solution is to reuse the Jaccard Index calculation of the "clinical signature extractor". This could have blind spots in the case where no public clinical data is available.

fhennig:
resolves https://github.com/cbg-ethz/WisePulse/issues/240

### Summary

Compute and display a Jaccard index column for mutations loaded from Nextclade tree collections, using the same clinical LAPIS queries as the computed signature mode. When no clinical sequences match the lineage (e.g. no public data available), the column is silently omitted.

### Screenshot

https://github.com/user-attachments/assets/3e040531-06ca-4c98-a255-53604142bd2c

### PR Checklist

- Tests ✅


fhennig:
Note to self, I haven't checked this yet.

fhennig:
I see a few issues:
- The Jaccard index isn't rounded to 2 digits, it looks weird - why isn't it rounded, is there code duplication?
- From a UX perspective, I find it confusing, where is this number coming from?
- If I Pick a variant like A.11, the column is just missing. There should be an info box

I'm thinking, maybe we could make an info box specifically titled 'Jaccard Index Information' and it would contain partially the information that is in the current box (how many clicincal seqs are there for the selected lineage?) I think that would help with the disconnect I'm now feeling

fhennig:
image

We have this now ... Hmm ... There are 100+ mutations, and we can't sort by relevancy, which is annoying.

i will try and see if we can implement the threshold at least.

... Hmmm with thresholding I don't see any relevant mutations anymore. very weird. I think I'll have to look at this differently.

---

I found these that seem with a "high" score:

image

In the other mode the scores are much higher. Something seems quite wrong with the computation.

I think I first need to reason about it myself: How is this computed? Since we're using two different data sets, it's easy to mix things up.

fhennig:
I found the "problem"! If we're computing Jaccard for XFG strictly (not XFG*) then the value is indeed very low. Not sure how to deal with this for the precomputed lineages. the non-* value is quite useless. But We shouldn't implicitly use the * without informing the user.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
```diff
@@ -161,23 +166,56 @@ function PredefinedSignature({
...
-                >
-                    <label htmlFor={newMutationsCheckBoxId} className='cursor-pointer pl-2'>
-                        Mutation not in parent
-                    </label>
-                </div>
             </div>
+            <CheckboxWithTooltip
+                checked={pageState.includeSublineagesForJaccard !== false}
+                onChange={(checked) => setPageState({ ...pageState, includeSublineagesForJaccard: checked })}
+                tooltip='Include sublineages when computing Jaccard index (appends * to the lineage)'
```
fengelniederhammer, line 189: Let's maybe also explain why one would include/exclude sublineages?

website/src/views/pageStateHandlers/WasapPageStateHandler.ts
```diff
@@ -154,6 +155,9 @@ export class WasapPageStateHandler implements PageStateHandler<WasapFilter> {
                     if (analysis.newMutationsOnly) {
                         setSearchFromString(search, 'newMutationsOnly', 'true');
                     }
+                    if (analysis.includeSublineagesForJaccard === false) {
+                        setSearchFromString(search, 'includeSublineagesForJaccard', 'false');
+                    }
```
fengelniederhammer, line 163: AI: In the predefined branch only `includeSublineagesForJaccard` is serialized; `minJaccard` (and the `timeFrame` used for the clinical Jaccard query) are not. Reproduced live: after toggling sublineages the URL is `...&collectionId=...&includeSublineagesForJaccard=false&` with no `minJaccard`/`timeFrame`, so a reload or shared link resets those to defaults. This breaks the "state in URL" contract for the new predefined Jaccard filtering. Serialize `minJaccard` and `timeFrame` in this branch too.

website/src/lapis/getMutations.ts
```diff
@@ -41,6 +41,43 @@ export async function getMutations(
...
+                if (!Object.hasOwn(totalCounts, mutation)) {
+                    throw new Error(
+                        `Data inconsistency: mutation ${mutation} observed in lineage but absent from global population query.`,
+                    );
+                }
+                return [mutation, count / (variantCount + totalCounts[mutation] - count)];
+            }),
+        );
+    });
+}
```
fengelniederhammer, line 61: AI: This duplicates the 3-query pattern and Jaccard formula of `getMutationsForVariant` (L96-115) — same `intersectionCounts`/`totalCounts`/`variantCount` fetch and the same `count / (variantCount + totalCounts[mutation] - count)` calculation. The two copies already diverge (this one throws on missing `totalCounts[mutation]`, the other doesn't). Consider extracting the shared Jaccard computation into one helper and having both call it, so the formula and inconsistency handling can't drift.

website/src/components/pageStateSelectors/wasap/filters/VariantExplorerFilter.tsx
fhennig: I think I'll leave it the way it is for now. I'm not sure if there is a good reason to turn this off; the checkbox is also half there just to communicate what we're doing.

We can change the tooltip later if we can come up with a better one.

website/src/lapis/getMutations.ts
```diff
@@ -41,6 +41,25 @@ export async function getMutations(
     );
 }
 
+/**
+ * Returns a map from mutation code to Jaccard index for all mutations observed in clinical
+ * sequences belonging to the given lineage. Mutations not observed in the lineage are absent
+ * from the map. Returns an empty map when no clinical sequences match the lineage filter.
```
fengelniederhammer: AI: This doc is now stale after the dedup refactor. It says "Returns an empty map when no clinical sequences match the lineage filter", but the `variantCount === 0` early-return was removed when this became a wrapper over `getMutationsForVariant`. With zero clinical sequences the function no longer short-circuits, so the "empty map" guarantee no longer holds as described. Please update the JSDoc to match the new behavior.

fengelniederhammer (approved):

fengelniederhammer (approved):

PR #1289, opened 2026-06-25
feat(dashboards): add collection tagging system

### LINKED ISSUE #1265: feat(backend): collection tagging
## Summary

Add a tagging system to collections on the backend: schema, DTO changes, and API updates.

Follow-up: #1266

## Database schema

Add a new `collection_tags` table (Flyway migration `V1.3__add_collection_tags.sql`):

```sql
CREATE TABLE collection_tags (
    collection_id INTEGER NOT NULL REFERENCES collections_table(id) ON DELETE CASCADE,
    tag           TEXT    NOT NULL CHECK (tag = lower(tag)),
    PRIMARY KEY (collection_id, tag)
);

CREATE INDEX idx_collection_tags_tag ON collection_tags(tag);
```

The `CHECK (tag = lower(tag))` constraint enforces lowercase at the DB level. The Kotlin layer should also call `.lowercase()` before inserting, so the constraint is never hit at runtime — if it is, that's a bug in the application code.

The PK `(collection_id, tag)` covers loading all tags for a given collection. The separate index on `tag` covers the filtering query (`WHERE tag = ?`), which would otherwise do a full table scan since `tag` is the second column of the PK.

**Why this schema (option A — junction table with free-text tags, no separate `tags` table):**
A separate `tags` table would give tags a canonical identity and make renaming atomic, but adds a join on every query and requires an upsert before linking. For free-text labels where renaming is not a current requirement, the simpler junction table is sufficient. Tags are stored normalised (lowercase) on write to avoid case-variant duplicates. Migrating to a separate `tags` table later is straightforward: create the table, backfill from distinct values in `collection_tags`, add a FK, done.

## Kotlin changes

- New `CollectionTagsTable` Exposed table object mirroring the schema above.
- `Collection` response DTO gains `tags: List<String>`.
- `CollectionRequest` gains `tags: List<String>` (defaults to empty list).
- `CollectionUpdate` gains `tags: List<String>?` (null = do not change).
- `CollectionModel`: load/save tags in all CRUD methods (replace-all on update); call `.lowercase()` on each tag before inserting.

## API changes

**`GET /collections`** — add repeatable `tags` query parameter.
- Semantics: AND — only collections that have **all** specified tags are returned.
- Example: `GET /collections?tags=flu&tags=europe` → collections tagged with both `flu` and `europe`.
- No tags param → behaviour unchanged (no filtering).

**`GET /collections/tags`** — new endpoint, returns all distinct tags currently in use across all users and organisms.
- Response: `{ "tags": ["europe", "flu", ...] }` (sorted).
- No auth required (read-only, non-sensitive).
- Useful for autocomplete on the frontend; kept intentionally simple for now.

fhennig:
resolves #1265 

### Summary

Implements tags on collections: database schema, CRUD support, tag
filtering on GET /collections, and a new GET /collections/tags endpoint
for frontend autocomplete.

### Screenshot

n/a

## PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
@@ -68,6 +68,7 @@ data class CollectionRequest(
 {
     "name": "Updated Collection Name",
     "description": "Updated description",
+    "tags": ["europe", "flu"],
     "variants": [
         {
             "type": "query",
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
@@ -14,6 +14,7 @@ import kotlin.time.Instant
     "organism": "covid",
     "description": "A collection of interesting variants",
     "variantCount": 1,
+    "tags": ["europe", "flu"],
     "createdAt": "2026-01-01T00:00:00Z",
     "updatedAt": "2026-01-02T00:00:00Z"
 }
@@ -28,6 +29,7 @@ data class Collection(
     val variantCount: Int,
     @JsonInclude(JsonInclude.Include.NON_NULL)
     val variants: List<Variant>?,
+    val tags: List<String>,
     val createdAt: Instant,
     val updatedAt: Instant,
 )
@@ -39,6 +41,7 @@ data class Collection(
     "name": "My Collection",
     "organism": "covid",
     "description": "A collection of interesting variants",
+    "tags": ["europe", "flu"],
     "variants": [
         {
             "type": "query",
@@ -55,6 +58,7 @@ data class CollectionRequest(
     val name: String,
     val organism: String,
     val description: String? = null,
+    val tags: List<String> = emptyList(),
     val variants: List<VariantRequest>,
 )
 
@@ -85,5 +89,9 @@ data class CollectionRequest(
 data class CollectionUpdate(
     val name: String? = null,
     val description: String? = null,
+    val tags: List<String>? = null,
     val variants: List<VariantUpdate>? = null,
 )
+
+@Schema(description = "Response containing all distinct tags in use")
+data class CollectionTagsResponse(val tags: List<String>)
--- a/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
+++ b/backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
@@ -33,6 +33,14 @@ import kotlin.time.Instant
 @Service
 @Transactional
 class CollectionModel(private val dashboardsConfig: DashboardsConfig, private val userModel: UserModel) {
+    /**
+     * The core function to fetch collections. Also fetches associated variants and tags (if desired).
+     *
+     * @param includeVariants whether to fetch variants as well. Quicker to omit if not needed.
+     * @param excludeSystemCollections whether to exclude collections belonging to the system user.
+     *   Useful when only community collections are desired, and filtering makes responses smaller.
+     * @param tags if provided, only collections containing all specified tags are returned.
+     */
     fun getCollections(
         userId: Long?,
         organism: String?,
@@ -48,32 +56,7 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
             dashboardsConfig.validateCollectionsEnabled(organism)
         }
 
-        var collectionConditions: Op<Boolean> = Op.TRUE
-        if (userId != null) {
-            collectionConditions = collectionConditions and (CollectionTable.ownedBy eq userId)
-        }
-        if (organism != null) {
-            collectionConditions = collectionConditions and (CollectionTable.organism eq organism)
-        }
-        if (excludeSystemCollections) {
-            val systemUserId = userModel.getSystemUserId()
-            if (systemUserId != null) {
-                collectionConditions = collectionConditions and (CollectionTable.ownedBy neq systemUserId)
-            }
-        }
-        if (!tags.isNullOrEmpty()) {
-            val distinctTags = tags.map { it.lowercase() }.distinct()
-            var matchingIds: Set<Long>? = null
-            for (tag in distinctTags) {
-                val idsWithTag = CollectionTagsTable
-                    .selectAll()
-                    .where { CollectionTagsTable.tag eq tag }
-                    .mapTo(mutableSetOf()) { it[CollectionTagsTable.collectionId].value }
-                matchingIds = if (matchingIds == null) idsWithTag else (matchingIds intersect idsWithTag)
-            }
-            collectionConditions = collectionConditions and
-                (CollectionTable.id inList (matchingIds?.toList() ?: emptyList()))
-        }
+        val collectionConditions = buildCollectionConditions(userId, organism, excludeSystemCollections, tags)
 
         val join = CollectionTable
             .join(VariantTable, JoinType.LEFT)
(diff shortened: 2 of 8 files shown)


backend/src/main/kotlin/org/genspectrum/dashboardsbackend/api/Collection.kt
```diff
@@ -55,6 +58,7 @@ data class CollectionRequest(
     val name: String,
     val organism: String,
     val description: String? = null,
+    val tags: List<String> = emptyList(),
```
fengelniederhammer, line 61: Should we also update the example here?

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
```diff
@@ -239,6 +272,37 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
...
+            val distinctTags = tags.map { it.lowercase() }.distinct()
+            val tagSubquery = CollectionTagsTable
+                .select(CollectionTagsTable.collectionId)
+                .where { CollectionTagsTable.tag inList distinctTags }
+                .groupBy(CollectionTagsTable.collectionId)
+                .having { Count(CollectionTagsTable.tag, distinct = true) eq distinctTags.size.toLong() }
+            conditions = conditions and (CollectionTable.id inSubQuery tagSubquery)
+        }
+        return conditions
+    }
```
fengelniederhammer, line 305:
```suggestion
    private fun buildCollectionConditions(
        userId: Long?,
        organism: String?,
        excludeSystemCollections: Boolean,
        tags: List<String>?,
    ): Op<Boolean> {
        val systemUserId = if (excludeSystemCollections) userModel.getSystemUserId() else null

        val conditions = buildList {
            if (userId != null) {
                add(CollectionTable.ownedBy eq userId)
            }
            if (organism != null) {
                add(CollectionTable.organism eq organism)
            }
            if (systemUserId != null) {
                add(CollectionTable.ownedBy neq systemUserId)
            }
            if (!tags.isNullOrEmpty()) {
                add(CollectionTable.id inSubQuery tagSubquery(tags))
            }
        }

        return conditions.fold(Op.TRUE as Op<Boolean>) { acc, condition -> acc and condition }
    }

    private fun tagSubquery(tags: List<String>): Query {
        val distinctTags = tags.map { it.lowercase() }.distinct()
        return CollectionTagsTable
            .select(CollectionTagsTable.collectionId)
            .where { CollectionTagsTable.tag inList distinctTags }
            .groupBy(CollectionTagsTable.collectionId)
            .having { Count(CollectionTagsTable.tag, distinct = true) eq distinctTags.size.toLong() }
    }
```

Not sure whether this is actually more readable, I just tried refactoring it into a `list.reduce` pattern. Maybe it could be used to build something more reusable, but also maybe not worth it now.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
```diff
@@ -44,25 +56,20 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
...
         return if (includeVariants) {
-            join.selectAll()
+            val tagsExpr = StringAgg(CollectionTagsTable.tag, orderBy = true)
+            val columns = CollectionTable.columns + VariantTable.columns + listOf(tagsExpr)
+
+            join.select(columns)
                 .where { collectionConditions }
+                .groupBy(CollectionTable.id, VariantTable.id)
+                .toList()
                 .groupBy { it[CollectionTable.id] }
```
fengelniederhammer, line 82: Do we still need this groupBy?

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
```diff
@@ -146,6 +159,14 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
             variantEntity
         }
 
+        val insertedTags = request.tags.map { it.lowercase() }.distinct().sorted()
```
fengelniederhammer: AI: risk: tag inserts are a per-row `INSERT` loop (N round-trips per create; same in `putCollection` L213). Use `CollectionTagsTable.batchInsert(insertedTags) { tag -> this[collectionId] = ...; this[CollectionTagsTable.tag] = tag }` to collapse to a single statement.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
```diff
@@ -112,13 +115,23 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
                         description = row[CollectionTable.description],
                         variantCount = row[countExpr].toInt(),
                         variants = null,
+                        tags = row[tagsExpr]?.split(",")?.sorted() ?: emptyList(),
```
fengelniederhammer: AI: nit: this path sorts tags in Kotlin (`.split(",").sorted()`) while the `includeVariants` path (L87) relies on SQL `ORDER BY` and does not re-sort. Same field, two ordering sources. Pick one — sort in SQL for both or in Kotlin for both — so behavior can't drift.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTagsTable.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val column: Column<String>,
+    private val distinct: Boolean = false,
+    private val orderBy: Boolean = false,
+) : Function<String?>(TextColumnType()) {
+    override fun toQueryBuilder(queryBuilder: QueryBuilder) = queryBuilder {
+        append("string_agg(")
+        if (distinct) append("DISTINCT ")
+        append(column)
+        append(", ','")
+        if (orderBy && !distinct) {
```
fengelniederhammer: AI: nit: `orderBy` is silently ignored when `distinct=true` (`orderBy && !distinct`). A caller asking for sorted+distinct gets unsorted output with no error. Postgres supports `string_agg(DISTINCT col, ',' ORDER BY col)` — either support the combination or document the limitation so callers don't rely on it.

backend/src/test/kotlin/org/genspectrum/dashboardsbackend/controller/CollectionsTagsTest.kt
fengelniederhammer, line 1: AI: test coverage gaps worth adding:
- collection with **multiple variants + multiple tags** in `includeVariants=true` — exercises the variant×tag join; would catch tag duplication if the aggregation shape changes.
- **duplicate tags in request** (e.g. `["flu","flu"]`) — verifies `.distinct()` before insert.
- **`getAllTags` on empty DB** — confirms empty-list response.
- **tag filter on the count path** (`includeVariants=false`) — all current filter tests pass `includeVariants=true`, so the non-variant branch's filtering is untested.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
fhennig: Yes, "it's what collapses the one-row-per-variant SQL results back into per-collection groups so we can build Collection objects with their full variants list." (Claude)

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTagsTable.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class JsonAgg(private val column: Column<String>) : Function<String?>(TextColumnType()) {
+    override fun toQueryBuilder(queryBuilder: QueryBuilder) = queryBuilder {
+        append("array_to_json(array_agg(DISTINCT ")
+        append(column)
+        append(" ORDER BY ")
+        append(column)
+        append("))")
+    }
+}
```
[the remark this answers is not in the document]
fengelniederhammer: AI: confirmed real bug, not theoretical. For an untagged collection the LEFT join yields rows with `tag = NULL`; `array_agg(DISTINCT tag)` over an all-NULL group returns `{NULL}`, so `array_to_json(...)` produces the literal string `"[null]"`. `readValue<List<String>>("[null]")` then deserializes to `listOf(null)` — the `?: emptyList()` guard never fires because the column value is non-null. Result: `GET /collections` returns `tags: [null]` for every untagged collection (both branches use `JsonAgg`, L94 + L125).

Fix: filter NULLs and coalesce, e.g. `array_to_json(coalesce(array_agg(DISTINCT tag) FILTER (WHERE tag IS NOT NULL), '{}'))`.

This is untested: the "creating collection without tags" test only asserts the POST response, which is built from `insertedTags` and never touches the aggregation. A `GET /collections` (or `getCollection`) assertion on a zero-tag collection would catch it.

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionModel.kt
```diff
@@ -44,25 +63,20 @@ class CollectionModel(private val dashboardsConfig: DashboardsConfig, private va
...
-        val join = CollectionTable.join(VariantTable, JoinType.LEFT)
+        val join = CollectionTable
+            .join(VariantTable, JoinType.LEFT)
+            .join(CollectionTagsTable, JoinType.LEFT, CollectionTable.id, CollectionTagsTable.collectionId)
 
         return if (includeVariants) {
-            join.selectAll()
+            val tagsExpr = JsonAgg(CollectionTagsTable.tag)
+            val columns = CollectionTable.columns + VariantTable.columns + listOf(tagsExpr)
+
```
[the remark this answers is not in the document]
fhennig: I've made a follow up for this: https://github.com/GenSpectrum/dashboards/issues/1295

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTable.kt
```diff
@@ -36,6 +39,11 @@ class CollectionEntity(id: EntityID<Long>) : LongEntity(id) {
 
     fun toCollection(): Collection {
         val variantList = variants.map { it.toVariant() }
+        val tagList = CollectionTagsTable
+            .selectAll()
+            .where { CollectionTagsTable.collectionId eq id.value }
+            .orderBy(CollectionTagsTable.tag to SortOrder.ASC)
+            .map { it[CollectionTagsTable.tag] }
```
fengelniederhammer, line 46: Can't we actually handle this automatically via some Entity magic? Like the `variants` field does?

fengelniederhammer (approved):

backend/src/main/kotlin/org/genspectrum/dashboardsbackend/model/collection/CollectionTable.kt
fhennig: Good idea, I checked. But the tags aren't entities (don't have an ID etc.) so it doesn't work for tags.

PR #1297, opened 2026-06-30
fix(website): fetch all collections so Official tab is not always empty

fhennig:
## Summary

- The collections overview API call hardcoded `excludeSystemCollections=true`, so system (official) collections were never fetched
- The **Official** filter tab, which shows collections owned by the system user, was therefore always empty
- Fix: remove the parameter so all collections are returned; the existing frontend `filterCollections` logic already handles splitting them into community/official/all


fengelniederhammer (approved):

fengelniederhammer:
@copilot fix the failing unit tests

PR #1308, opened 2026-07-01
feat(website): add GET /users/me endpoint for authenticated user lookup

fhennig:
### Summary

Adds a new `/api/users/me` endpoint.

I want to use this, so the seeder can - based on the given API key - find out its own user ID. It can then use this to fetch its own collections.

## PR Checklist


fhennig:
From the backend perspective this is a little bit stupid ... Should this be a frontend only thing?

fengelniederhammer (commented):
Let's change the PR title to `feat(website)` and add an e2e (Playwright) test for the new endpoint?

fengelniederhammer (approved):
Nice :)

PR #1309, opened 2026-07-01
fix(website): remove incorrect excludeSystemCollections param from tag filter test mocks

fhennig:
## Summary

- The 5 new `tag filter` tests in `CollectionsOverview.browser.spec.tsx` were all timing out because their mocks were set up with `excludeSystemCollections: true` as a required request param
- The component does all community/official/all filtering client-side and never sends `excludeSystemCollections` in its API request
- The `resolver` in `routeMocker.ts` does strict param matching, so the mock never matched → returned 400 → component had no data
- Fix: remove the spurious `true` 4th argument from all 5 `mockGetCollectionSummaries` calls in the tag filter describe block

## Test plan

- [x] All 11 `CollectionsOverview` browser tests now pass locally (previously 5 failed)


fhennig:
Not sure how this slipped through

fengelniederhammer (approved):

PR #1331, opened 2026-07-09
docs(agents): specify how we want string union types to be defined in AGENTS.md

fhennig:
### Summary

We want a `const` object so we don't have to use literal strings in the code.


fengelniederhammer (approved):
👍

PR #1332, opened 2026-07-09
test(wasap): add tests for fetchWasapPageData

fhennig:
## Summary

- Adds `useWasapPageData.spec.ts` with 17 tests covering all 5 fetch modes: manual, resistance, variant (computed + predefined signatures), untracked (custom + predefined exclude sets), and covSpectrumCollection — including enabled, disabled, and error cases
- Exports `fetchWasapPageData` for direct testing (avoids React Query `renderHook` setup)
- Adds `BackendRouteMocker.mockGetCollection` to `routeMocker.ts`
- Converts `SignatureType`, `WasapAnalysisMode`, and `ExcludeSetName` from bare string unions to the const-object pattern (matching `VARIANT_TIME_FRAME`), and adds `SEQUENCE_TYPE` for the external `SequenceType`

## Test plan

- [x] `npx vitest run src/components/views/wasap/useWasapPageData.spec.ts` — all 17 tests pass
- [x] `npx tsc --noEmit` — no errors

### DIFF
--- a/website/src/components/views/wasap/useWasapPageData.spec.ts
+++ b/website/src/components/views/wasap/useWasapPageData.spec.ts
@@ -1,0 +1,525 @@
+import dayjs from 'dayjs';
+import { http } from 'msw';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+import { fetchWasapPageData, getLapisFilterForTimeFrame } from './useWasapPageData.ts';
+import {
+    EXCLUDE_SET_NAME,
+    SEQUENCE_TYPE,
+    SIGNATURE_TYPE,
+    VARIANT_TIME_FRAME,
+    WASAP_ANALYSIS_MODE,
+    type WasapPageConfig,
+} from './wasapPageConfig.ts';
+import { DUMMY_BACKEND_URL, DUMMY_LAPIS_URL } from '../../../../routeMocker.ts';
+import { astroApiRouteMocker, backendRouteMocker, lapisRouteMocker, testServer } from '../../../../vitest.setup.ts';
+import type { Collection } from '../../../types/Collection.ts';
+
+vi.mock('../../../backendApi/backendService.ts', async (importOriginal) => {
+    const mod = await importOriginal<typeof import('../../../backendApi/backendService.ts')>();
+    return {
+        ...mod,
+        getBackendServiceForClientside: () => new mod.BackendService(DUMMY_BACKEND_URL),
+    };
+});
+
+const DUMMY_COV_SPECTRUM_URL = 'http://cov-spectrum.dummy/api/v2';
+
+// these fields have no effect on data fetching, but need to be present to have a correct type.
+const unusedBaseConfigFields = {
+    internalName: '',
+    name: '',
+    path: '',
+    description: '',
+    linkTemplate: { nucleotideMutation: '', aminoAcidMutation: '' },
+    samplingDateField: '',
+    locationNameField: '',
+    defaultLocationName: '',
+    browseDataUrl: '',
+    browseDataDescription: '',
+};
+
+const baseConfigFields: WasapPageConfig = {
+    ...unusedBaseConfigFields,
+    lapisBaseUrl: DUMMY_LAPIS_URL,
+};
+
+describe('fetchWasapPageData', () => {
+    beforeEach(() => {
+        astroApiRouteMocker.mockLog();
+    });
+
+    describe('manual mode', () => {
+        const config = {
+            ...baseConfigFields,
+            manualAnalysisModeEnabled: true as const,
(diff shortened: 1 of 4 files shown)


website/src/components/views/wasap/useWasapPageData.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+                    {
+                        type: 'filterObject',
+                        id: 1,
+                        collectionId: 1,
+                        name: 'Nucleotide substitutions',
+                        description: null,
+                        filterObject: { nucleotideMutations: ['A123T', 'G456C'] },
+                    },
+                ],
+            } as unknown as Collection);
```
fengelniederhammer, line 195: AI: nit: `as unknown as Collection` double-cast bypasses type checking. The mock looks structurally complete — type it as `Collection` directly so a future `Collection` type change is caught by this test.

website/src/components/views/wasap/useWasapPageData.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+                    collectionId: 1,
+                    includeSublineagesForJaccard: true,
+                },
+            );
+
+            expect(result).toEqual({
+                type: 'mutations',
+                displayMutations: ['A123T', 'G456C'],
+                lineageForJaccard: 'XEC*',
+                // TODO - are we sure this makes sense? we don't have the custom column here
```
fengelniederhammer: AI: nit: committed TODO questioning the assertion itself ("are we sure this makes sense?"). Resolve the doubt and add the missing non-empty-jaccard / custom-column test, or drop the TODO before merge.

website/src/components/views/wasap/useWasapPageData.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+                    { mode: WASAP_ANALYSIS_MODE.covSpectrumCollection, collectionId: undefined },
+                ),
+            ).rejects.toThrow('No collection selected');
+        });
+    });
+});
+
+describe('getLapisFilterForTimeFrame', () => {
+    test('"all" returns an empty filter', () => {
+        expect(getLapisFilterForTimeFrame('all', 'date')).toEqual({});
```
fengelniederhammer: AI: nit: these tests pass raw `'all'`/`'6months'`/`'3months'` while the file already imports `VARIANT_TIME_FRAME`. Use `VARIANT_TIME_FRAME.*` for consistency with the rest of the file.

website/src/components/views/wasap/useWasapPageData.spec.ts
fengelniederhammer, line 1: AI: coverage gaps (ok as follow-up, not blocking):
- predefined variant: variant-not-found throw (src L137), non-filterObject throw (L140), aminoAcid + `newMutationsOnly` name selection (L130-134), non-empty jaccard filter branch (L161-178).
- covSpectrumCollection: `detailedMutations` query type (L262), empty-query `'Variant is empty.'` (L267), `validateGenomeOnly` non-genome rejection (L306).
- untracked: `variantsToExclude === undefined` → `[]` (L207).
- `mockGetCollection` non-200 `statusCode` / error path never exercised.

fengelniederhammer (commented):
A bunch of nits, looks good otherwise 👍

website/src/components/views/wasap/useWasapPageData.spec.ts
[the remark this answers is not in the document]
fhennig: No, it doesn't work:

⏺ The problem is in filterObjectSchema at Collection.ts:10. The .catchall(z.string()) adds an index
  signature { [k: string]: string } to the inferred type. TypeScript then requires all named
  properties to also satisfy that index signature — but nucleotideMutations?: string[] has type
  string[], which is not assignable to string.

  Zod's inferred type becomes this impossible intersection:

  {
      nucleotideMutations?: string[] | undefined;
      // ...
  } & { [k: string]: string }   // ← string[] is not string

  So { nucleotideMutations: ['A123T', 'G456C'] } can't be cast directly to Collection.

  The fix in the test is the standard TypeScript escape hatch for this situation — cast through
  unknown first:

  } as unknown as Collection);

  The underlying schema design is a bit odd (the catchall is meant to capture extra lineage fields
  like pangoLineage: 'XEC*' — scalars — but the known fields are arrays, creating the
  contradiction). If you want to fix the type itself rather than the test, you'd change the
  catchall to z.union([z.string(), z.array(z.string())]). But the cast in the test is the minimal
  fix

website/src/components/views/wasap/useWasapPageData.spec.ts
fhennig: yeah, we don't need perfect coverage of everything.

fengelniederhammer (approved):

